PR #2, opened 2017-09-29
Fix tests when not running under Windows

JosephWoodward:
Also added Rider and VsCode to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -200,6 +200,8 @@ ClientBin/
 *.pfx
 *.publishsettings
 orleans.codegen.cs
+.idea
+.vscode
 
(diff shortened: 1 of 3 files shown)


JosephWoodward:
There may be a more elegant way of doing this, if there is then I'm all ears :)

daveaglick:
Awesome, thanks a ton for this! The project is specifically designed to address cross-platform use cases, so it's nice that the tests will actually run :smile: 

Did you have any luck getting it to work for your IL Viewer?

JosephWoodward:
Yeah, this does just the job 👍  Next is trying to resolve external dependencies which is proving to be a challenge (which may be worth including in Buildalyzer once I've figured it out?)

PR #6, opened 2017-10-09
GH1: Identify SDK project based on import statement.

### LINKED ISSUE #1: Identify SDK projects by import statements
Turns out you can also identify SDK projects via imports and there may not be a root SDK attribute:
https://github.com/Microsoft/msbuild/issues/1493

Right now we just look for the SDK attribute on the root element in the project file. Add support for locating SDK import elements if the SDK attribute isn't found.

mholo65:
- Fixes #1


daveaglick:
Thanks a bunch! This little project is starting to come together :smile:. I went ahead and just rebased and merged this manually since it was just the solution file that had a conflict.

Figured out the problems I was having locally were due to the VS update changing my test settings to a x86 host which apparently caused the in-proc MSBuild to freak out a little. All is good.

PR #9, opened 2017-10-14
Hide full framework tests when on .NET Core

JosephWoodward:
These tests currently fail against .NET Core, as they're full framework tests this will only discover them when compiling against the full framework.


daveaglick:
:+1:

PR #34, opened 2018-02-06
increase timeout to get the dotnet --info output from 1 to 4 seconds

Mpdreamz:
Completely anecdotal but this fixed our `TeamCity` builds that were failing on this.


daveaglick:
Thanks - turns out I actually just started hitting issues on getting the output in AppVeyor recently too so I'm hoping this does the trick.

PR #35, opened 2018-02-06
allow for opt-in inclusion of projects in sln

Mpdreamz:


src/Buildalyzer/AnalyzerManager.cs
```diff
@@ -63,12 +68,13 @@ public AnalyzerManager(string solutionFilePath, TextWriter logWriter, LoggerVerb
...
                 SolutionDirectory = Path.GetDirectoryName(solutionFilePath);
-                GetProjectsInSolution(solutionFilePath);
+                GetProjectsInSolution(solutionFilePath, projects);
             }
         }
 
-        private void GetProjectsInSolution(string solutionFilePath)
+        private void GetProjectsInSolution(string solutionFilePath, string[] projects = null)
         {
+            projects = projects ?? new string[] { };
```
daveaglick: Instead of allocating an empty array and then checking
```
if (projects.Length > 0 && !projects.Contains(project.ProjectName))
```
further down, I'm wondering if it would be better to skip the empty array and make the check
```
if(projects?.Contains(project.ProjectName) ?? false)
```

Thoughts?

tests/FrameworkTests/FrameworkTests.csproj
```diff
@@ -50,28 +50,28 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.Extensions.Configuration.Abstractions, Version=2.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60, processorArchitecture=MSIL">
-      <HintPath>..\packages\Microsoft.Extensions.Configuration.Abstractions.2.0.0\lib\netstandard2.0\Microsoft.Extensions.Configuration.Abstractions.dll</HintPath>
+      <HintPath>..\..\packages\Microsoft.Extensions.Configuration.Abstractions.2.0.0\lib\netstandard2.0\Microsoft.Extensions.Configuration.Abstractions.dll</HintPath>
```
daveaglick: Any idea why the hint paths got changed?

src/Buildalyzer/AnalyzerManager.cs
Mpdreamz: Much better! will amend

tests/FrameworkTests/FrameworkTests.csproj
Mpdreamz: I manually edited this because restore put the packages in the root dir which is two folder up, in order to compile on a clean check out. Happy to leave this out of the PR if needed.

src/Buildalyzer/AnalyzerManager.cs
Mpdreamz: Went with another variation in 8a3cf6b, let me know what you think :)

tests/FrameworkTests/FrameworkTests.csproj
daveaglick: Nope - leave it. Was just wondering what happened. I’d much rather it compile from check-out, not sure why my setup was different.

daveaglick:
Thanks! Going to try and get a release out this afternoon with these changes.

PR #36, opened 2018-02-06
Feature/tweak project support

Mpdreamz:
This is an extension of #35 and allows an action to be injected that can modify the found project files. 

We need this for [elastic/elasticsearch-net](http://github.com/elastic/elasticsearch-net) documentation generation to rewrite certain `<Import/>`'s.


daveaglick:
Totally on board with this, thanks! Since it builds on #35, I'll wait for that one to get merged before bringing in this one.

Issue #33, opened 2018-02-04
PackageReference assemblies not returned by GetReferences() for full framework projects


jmezach:
It seems that GetReferences() on ProjectAnalyzer doesn't return assemblies from NuGet packages if PackageReference is being used instead of packages.config. This causes compilation errors in Roslyn workspaces.

Interestingly this only seems to be a problem when the host is .NET Core. It works just fine when the host is .NET Framework. I added some tests in this [commit](https://github.com/jmezach/Buildalyzer/commit/1cb084b5c640e629df574dcf3e5d7b11952145bf) that show this behaviour. I also see errors in the test output which are very similar to #29.

daveaglick:
I suspect the problem here is that the NuGet targets that know how to handle `PackageReference` are only included with the .NET Core SDK. It's possible that Visual Studio does some magic to locate those targets even for classic .NET project files. I also see an interesting comment to the docs [here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files):

```
PackageReference can be used with non-NETCore projects if the project system supports it--so the real answer depends on Unity itself. As for NuGet, it checks if the project supports VSProject4 to determine if it can read/write PackageReferences
```

This also suggests there's some kind of connection to the VS project system.

Are you able to submit your tests as a PR and I'll bring them in. Once we've got the failing tests merged we can take it from there.

jmezach:
Sure thing. Created #37.

jmezach:
Looks like this got worse. I'm now completely unable to use this library to load up a full framework solution which has project to project references. Not sure what changed, but I think it has to do with a recent Visual Studio update. Any progress on this issue?

daveaglick:
Not with this issue specifically, but I'll be pushing a new release within the hour that addresses the broader problems you described. Recently, every time VS or the .NET Core SDK updates Buildalyzer breaks because the MSBuild APIs that it uses aren't compatible with the newest MSBuild stuff in the SDK.

daveaglick:
Been chasing this down and I'm pretty sure at this point that the .NET Core SDK can't build legacy project files that use `PackageReference` (even though it _can_ build legacy project files). The problem is a missing `ResolveNugetPackageAssets` target:

![2018-06-22_13h40_02](https://user-images.githubusercontent.com/1020407/41791326-6f10c6fe-7623-11e8-9fb8-b4b21bc82c3f.png)

[...]

PR #37, opened 2018-02-17
Add tests for PackageReference projects

jmezach:
This PR adds tests for projects that use PackageReference instead of packages.config. Some of them are failing as noted in daveaglick/Buildalyzer#33.


daveaglick:
Thanks! I may end up commenting these out in master and then uncommenting them when I work on the issue (so don't panic when you see I've turned the new tests off - I'm working on them locally).

Issue #38, opened 2018-02-18
Outputs get cleaned


dfederm:
It seems to kind of defeat the purpose of a design-time build to do a clean beforehand. If the project were previously built via normal methods, the outputs will disappear after doing a design-time build.

It seems like at the very least this should be a flag,

daveaglick:
IIRC there was a reason why I ended up forcing a clean - something about .NET Core builds getting confused if the obj directory was hanging around from a previous build. I'd need to go back and take a closer look to be sure.

I'd be fine making it a switch though - the default behavior would need to stay the same so we don't break anyone.

dfederm:
Thanks for the context. PR with the addition of a flag is out :)

PR #40, opened 2018-02-22
Allow the Clean before Compile to be turned off

dfederm:
Fixes: #38

### DIFF
--- a/src/Buildalyzer/AnalyzerManager.cs
+++ b/src/Buildalyzer/AnalyzerManager.cs
@@ -1,16 +1,12 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.IO;
-using System.Runtime.Loader;
-using System.Text;
-using System.Xml;
-using Microsoft.Build.Framework;
-using Microsoft.Extensions.Logging;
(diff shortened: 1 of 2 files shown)


dlech:
This would be useful to me as well.

This is totally a matter of personal preference, but I'm wondering if it would be better to introduce a new AnalyzerManagerOptions class for passing options instead of adding more parameters to the AnalyzerManager constructor for each option. The AnalyzerManagerOptions would just be a class with read/write properties for each option. This way, to add a new option, you just add a new property to AnalyzerManagerOptions rather than ending up with a really long parameter list.

Example:

```C#
var manager = new AnalyzerManager(new AnalyzerManagerOptions {
    LoggerFactory = MyLoggerFactory,
    LoggerVerbosity = LoggerVerbosity.Quiet,
    CleanBeforeCompile = false
});
```

dfederm:
@dlech Completely agree about an options model, as the params are starting to get unruly. I'll leave that design decision to @daveaglick though :)

daveaglick:
Totally agree the constructor is doing too much, opened #44 to take a look at that after I merge this PR (thanks, BTW!). Will be looking to merge this in the next few days.

PR #41, opened 2018-02-22
Allow for custom build environments

dfederm:
I have a scenario where a custom build environment is used (effectively an xcopy install of MsBuild), so this change allows a `BuildEnvironment` to be explicitly provided. I ended up refactoring `BuildEnvironment` to be just a POCO and `EnvironmentFactory` contains all the logic previously contained in the subclasses and still provides good defaults.


daveaglick:
I like this change a whole lot, nice work. Out of curiosity, which team are you on?

dfederm:
@daveaglick I work on [CloudBuild](https://www.microsoft.com/en-us/research/publication/cloudbuild-microsofts-distributed-and-caching-build-service/) at Microsoft specifically on the build engine part of it. I share the same manager as the MsBuild folks as well, so this project interested me on a few levels :)

PR #45, opened 2018-03-27
Changed MSBuild Project construction to set project file path

jirikopecky:
This resolves usage of properties like MSBuildThisFileDirectory in project files.

I had actually taken the code from Roslyn MSBuild Workspaces project:
https://github.com/dotnet/roslyn/blob/master/src/Workspaces/Core/Desktop/Workspace/MSBuild/ProjectFile/ProjectFileLoader.cs#L54


daveaglick:
Thanks a ton for this! A new release will go out soon (possibly today).

PR #59, opened 2018-06-23
Adds ability to set Global Properties at the AnalyzerManager level

### LINKED ISSUE #52: Suggestion: Allow solution (project) global properties to be controlled via AnalyzerManager
I have a situation in which I'm loading a Web Application project using .NET Core. The following lines in that project type are problematic:

```
<VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>

<Import Project="$(VSToolsPath)\WebApplications\Microsoft.WebApplication.targets" Condition="'$(VSToolsPath)' != ''" />
```

Since `VSToolsPath` is undefined and `MSBuildExtensionsPath32` is set to `C:\Program Files\dotnet\sdk\<version>`, MSBuild ends up trying to load `C:\Program Files\dotnet\sdk\<version>\Microsoft\VisualStudio\v$(VisualStudioVersion)\WebApplications\Microsoft.WebApplication.targets`, which obviously doesn't exist (that file resides in the Visual Studio install directory). This causes `AnalyzerManager.GetWorkspace()` to blow up.

I've (very nastily) worked around this by calling `Environment.SetEnvironmentVariable("VSToolsPath", <path_to_necessary_directory>)`, but the issue would be totally avoided if I could somehow pass these values to `AnalyzerManager` when constructing it or calling `GetWorkspace()`. Unfortunately, Buildalyzer currently doesn't expose the global/environment variables it sets to allow for mutation, so that's not an option - but I propose it should be! :)

dfederm:
Adds ability to set Global Properties at the `AnalyzerManager` level. Also allows setting environment variables and fixes global property leak when sharing a `BuildEnvironment`.

Closes #52

### DIFF
--- a/src/Buildalyzer/ProjectAnalyzer.cs
+++ b/src/Buildalyzer/ProjectAnalyzer.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.IO;
 using System.Linq;
 using System.Xml;
@@ -26,6 +27,10 @@ namespace Buildalyzer
         private Project _project = null;
         private ProjectInstance _compiledProject = null;
 
+        // Project-specific global properties and environment variables
+        private readonly Dictionary<string, string> _globalProperties = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, string> _environmentVariables = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+
         public AnalyzerManager Manager { get; }
 
         public string ProjectFilePath { get; }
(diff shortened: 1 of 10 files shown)


src/Buildalyzer/ProjectAnalyzer.cs
```diff
@@ -186,16 +202,66 @@ public void SetGlobalProperty(string key, string value)
...
+        public void SetEnvironmentVariable(string key, string value)
+        {
+            if (_project != null)
+            {
+                throw new InvalidOperationException("Can not change environment variables once project has been loaded");
+            }
+            _environmentVariables[key] = value;
+        }
+
+        // Note the order of precedence (from least to most)
```
dfederm: Basically the `BuildEnvironment` variables are applied first, then the `AnalyzerManager` ones are applied (which may override the `BuildEnvironment` ones), and finally the `ProjectAnalyzer` ones are applies, which can override the other two.

daveaglick:
I love everything about this PR - very nice work!

PR #61, opened 2018-07-08
Add readme badges for NuGet package

dfederm:
It looks like:
image

Or, see it in action:
https://github.com/dfederm/Buildalyzer/tree/dfederm/readme-badges


daveaglick:
Thanks!

PR #69, opened 2018-08-03
Upgrade depending packages [account for transitivities]

ltcmelo:
There's a dependency transitivity issue (e.g., Roslyn and logging) that "forces" me to upgrade these packages. Note that my commits from the other cross-platform PR are also here (feel free to cherry-pick).


ltcmelo:
BTW, I had to remove DependencyModel because there doesn't seem to be an upgraded one available (luckily you're not using it).

daveaglick:
Looking at this now. Even though DependencyModel isn't directly used by Buildalyzer, MSBuild needs it available during the API use (not sure why it's not available in the MSBuild folders). Not a problem, I added it back in. Continuing to test this PR and will merge when it looks like I've got the kinks worked out.

daveaglick:
Thanks! Did similar updates for the other projects in the solution and now everything looks good.

PR #74, opened 2018-08-27
Fix an error with file without BOM like it can happened with Assembly…

bhugot:
…Info.cs when it's generated


daveaglick:
Nice catch, thanks!

PR #82, opened 2018-10-01
Add PackageReferences

mholo65:
- Add property for quickly accessing package references.

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -82,6 +82,11 @@ namespace Buildalyzer
                     Path.Combine(Path.GetDirectoryName(Analyzer.ProjectFile.Path), x.ItemSpec)))
                 : Array.Empty<string>();
 
+        public IEnumerable<(string package, string version)> PackageReferences =>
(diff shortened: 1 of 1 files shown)


daveaglick:
Thanks!

In general not a huge fan of exposing tuples as return values from a public API.

Do any of these alternatives sound better to you:
* `IEnumerable<PackageReference>` (a new struct with `Package` and `Version` properties)
* `IEnumerable<KeyValuePair<string, string>>`
* `IReadOnlyDictionary<string, string>` (key = package name, value = version)

I think I like returning a dictionary for this, but could be talked into something else.

BTW - hopefully CI will light up green on next trigger. Been working on that, but don't worry about it if it's still red.

mholo65:
I like the first one best:
> IEnumerable<PackageReference> (a new struct with Package and Version properties)

I'll update the PR.

mholo65:
Sorry, was to quick on the "Comment" button 😄 Dictionary also has it advantages, and would be possible to add other properties dynamically. I think I'll lean into that.

daveaglick:
Went ahead and merged this - changed it to a `IReadOnlyDictionary<string, IReadOnlyDictionary<string, string>>` so that you can find package references by package ID but also get all the metadata including version.

PR #83, opened 2018-10-05
avoid exception when adding specific property in EnvironmentOption

itn3000:
This PR fixes ArgumentException when following key exists in EnvironmentOption.EnvironmentVariables.

* MSBuildSDKsPath
* MSBuildExtensionsPath


itn3000:
appveyor tests are failed? In my environment(ubuntu 18.04 amd64), all tests are successed.

daveaglick:
Thanks for the fix - good find.

Don't worry too much about the CI failure. It's intermittent and only on the Linux build - haven't quite been able to track it down yet.

PR #84, opened 2018-10-05
add dotnet.exe setting variable for zip deployed SDK

itn3000:
# Motivation

when I want to use daily SDK build, I use zip-deployed SDK(currently, dotnet-sdk-3.0.100-*).
Currently, I must prepend custom sdk's dotnet.exe directory path to  "PATH" env.
Because environment variable may have session-global effect, I want to avoid setting PATH env.

# Suggestion

adding "DotNetExePath" to EnvironmentOptions, and then use it in build.


daveaglick:
Thanks! Very nice work.

PR #90, opened 2018-10-28
Improve Process Output/Error processing

jonstelly:
The way ProcessRunner was used, there was a race condition where the process could exit, WaitForExit would complete, and the process would be disposed of before all of the output was read.  There was a check in DotnetPathResolver for 0 lines, but I saw many cases where there was at least 1 line of output returned, but not all of them.

Switched from the event-based output/error processing to use ReadToEnd() on the Output and Error streams, encapsulated the Process more.  This code could be cleaned up further but I wanted to minimize change a bit.


daveaglick:
This PR looks like it's got stuff from the other PRs mixed in? Any chance you could refactor it to just include the process output work? If not, no big deal - I can fix it up on my end.

PR #91, opened 2018-10-29
Parallelize project builds during workspace creation

jonstelly:
Parallelize the project builds when creating a workspace.  This massively reduces the time to generate the workspace for medium to large solutions.

I'm unsure about thread safety issues here and would welcome feedback but I did test this pretty extensively by loading a medium sized solution and generating a workspace 100 times in a loop.  I didn't see any errors and saw consistent structure in the generated workspace.

NOTE: I tested this on a branch that also had https://github.com/daveaglick/Buildalyzer/pull/90, otherwise I ran into warnings because the dotnet --info output wasn't being fully parsed.


daveaglick:
Thanks! I _think_ the parallel builds will be okay - I tried to minimize threading issues while writing everything in the first place. Your anecdotal evidence also helps. I'm okay pulling this in and fixing stuff if it breaks.

PR #104, opened 2019-03-14
[WIP] read binlogs from solution and find latest msbuild 

colombod:
msbuild cannot be located unless using specific SKUS of VS 2017, now all vs fodler are scanned.

when loading a binlog from a build performed on a solution there are 0 results

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -108,16 +108,23 @@ namespace Buildalyzer
 
         internal void ProcessProject(ProjectStartedEventArgs e)
         {
-            // Add properties
-            foreach (DictionaryEntry entry in e.Properties.Cast<DictionaryEntry>())
+            if (e.Properties != null)
             {
-                _properties[entry.Key.ToString()] = entry.Value.ToString();
+                // Add properties
+                foreach (DictionaryEntry entry in e.Properties.Cast<DictionaryEntry>())
+                {
+                    _properties[entry.Key.ToString()] = entry.Value.ToString();
+                }
             }
 
-            // Add items
-            foreach (IGrouping<string, DictionaryEntry> itemGroup in e.Items.Cast<DictionaryEntry>().GroupBy(x => x.Key.ToString()))
+            if (e.Items != null)
             {
-                _items[itemGroup.Key] = itemGroup.Select(x => new ProjectItem((ITaskItem)x.Value)).ToArray();
+                // Add items
+                foreach (IGrouping<string, DictionaryEntry> itemGroup in e.Items.Cast<DictionaryEntry>()
+                    .GroupBy(x => x.Key.ToString()))
+                {
+                    _items[itemGroup.Key] = itemGroup.Select(x => new ProjectItem((ITaskItem)x.Value)).ToArray();
+                }
             }
         }
 
--- a/src/Buildalyzer/Logging/EventProcessor.cs
+++ b/src/Buildalyzer/Logging/EventProcessor.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Build.Framework;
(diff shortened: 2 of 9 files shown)


tests/Buildalyzer.Tests/Integration/SolutionWithProjectsFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+                processRunner.Start();
+                processRunner.WaitForExit(60000);
+            }
+
+            using (ProcessRunner processRunner = new ProcessRunner("dotnet", $"build {solutionPath} /bl:{binlogPath}", Path.GetDirectoryName(solutionPath), environmentVariables, null))
+            {
+                processRunner.Start();
+                processRunner.WaitForExit(60000);
+            }
+            */
```
daveaglick: Any particular reason the binlog generation code here is commented out? Without it the tests obviously fail on a new checkout.

src/Buildalyzer/AnalyzerResult.cs
```diff
@@ -108,16 +108,23 @@ internal AnalyzerResult(string projectFilePath, AnalyzerManager manager, Project
 
         internal void ProcessProject(ProjectStartedEventArgs e)
         {
-            // Add properties
-            foreach (DictionaryEntry entry in e.Properties.Cast<DictionaryEntry>())
+            if (e.Properties != null)
```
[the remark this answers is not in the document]
colombod: Yes I hit that with the tests I have added. I still need to work thisnproperly

src/Buildalyzer/Logging/EventProcessor.cs
```diff
@@ -60,24 +67,43 @@ public EventProcessor(AnalyzerManager manager, ProjectAnalyzer analyzer, IEnumer
...
+            if (_projectFilePaths.Contains(normalizedProjectFilePath))
             {
                 // Get the TFM for this project
-                string tfm = e.Properties?.Cast<DictionaryEntry>()
-                    .FirstOrDefault(x => string.Equals(x.Key.ToString(), "TargetFrameworkMoniker", StringComparison.OrdinalIgnoreCase)).Value as string;
+                string tfm = GetTargetFramework(e);
                 if (!string.IsNullOrWhiteSpace(tfm))
                 {
-                    if (!_results.TryGetValue(tfm, out AnalyzerResult result))
+                    string key = $"{normalizedProjectFilePath}:{tfm}";
```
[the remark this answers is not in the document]
colombod: This exactly what i’d Like to design and discuss with you once back

tests/Buildalyzer.Tests/Integration/SolutionWithProjectsFixture.cs
colombod: Is a work in progress and will execute that code. I just checked this is for you to see where I am trying to go with the changes

daveaglick:
@colombod what are your thoughts on this PR? I'd like to do another release soon now that VS 2019 is out and I think it makes sense to include this solution work.

My thinking right now is that adding an extra `AnalyzerProjectResults` collection to gather results for each project is the way to go. I can do that plumbing.

Is there anything else you think needs to be done from your end before I start working the client API and merging this in?

src/Buildalyzer/Logging/EventProcessor.cs
```diff
@@ -89,9 +115,29 @@ private void ProjectStarted(object sender, ProjectStartedEventArgs e)
             _currentResult.Push(null);
         }
 
+        private string GetTargetFramework(ProjectStartedEventArgs e)
+        {
+            string normalizedProjectFilePath = AnalyzerManager.NormalizePath(e.ProjectFile);
+            string tfm = null;
+            if (!_targets.TryGetValue(normalizedProjectFilePath, out tfm))
```
daveaglick: This part is confusing me. If I'm reading it correctly, we're caching the first TFM we see for each project file and always returning that as the TFM for that project file? That doesn't seem right... We want to always return the current project's TFM, right - that's why the `_results` key was changed to include TFM in the first place.

PR #106, opened 2019-03-31
change msbuild discovery path logic

colombod:
Change to logic for msbuild.exe  discovery


daveaglick:
Thanks for pulling this out separate from the WIP solution loading - makes it easier to merge while figuring out the solution part.

PR #115, opened 2019-07-23
disable env var "COREHOST_TRACE" for .NET Core

sapsari:
When working on .NET Core, if environment variable "COREHOST_TRACE" is set to 1, huge amounts of debug text is printed within dotnet commands.

This debug text makes Buildalyzer freeze (wait longer than a minute to fetch .NET SDK path, in method DotnetPathResolver.GetInfo)

Disabling this environment variable on dotnet command processes makes Buildalyzer run on expected speeds again.


daveaglick:
Awesome find, thanks! I can imagine someone setting the variable, forgetting it's set, and then running into problems when they try to run an app that uses Buildalyzer without even realizing what's wrong.

PR #120, opened 2019-10-25
Fix for finding the default SDK in some scenarios

patriksvensson:
When trying to find the default SDK and there isn't a global.json the default SDK finder sometimes throw a ArgumentOutOfRange exception because the line breaks between sections in the output from `dotnet --info` is missing for some reason.

This commit fixes one of those scenarios while keeping the original way of finding the SDK intact.

### DIFF
--- a/src/Buildalyzer/Environment/DotnetPathResolver.cs
+++ b/src/Buildalyzer/Environment/DotnetPathResolver.cs
@@ -109,11 +109,26 @@ namespace Buildalyzer.Environment
             {
                 return null;
             }
+
             index++;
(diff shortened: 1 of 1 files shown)


patriksvensson:
Whoopsie. Failing tests. Didn't see a test project but I'll make sure they pass.

daveaglick:
Don’t spend too much time on the tests if it doesn’t look like something introduced from your change - the integration test suite can be flaky. I’ll take a closer look soon.

daveaglick:
Thanks, and sorry for the delay merging this in!

PR #148, opened 2020-10-05
The SolutionDir property is different from the MSBuild output 

laurenceSaes:
The SolutionDir property does not end with a directory separator. This is different than how MSBuild sets this property:


laurenceSaes:
See also:

https://docs.microsoft.com/en-us/cpp/build/reference/common-macros-for-build-commands-and-properties?view=vs-2019

```
$(SolutionDir) | The directory of the solution (defined as drive + path); includes the trailing backslash '\'. Defined only when building a solution in the IDE.
-- | --
```

daveaglick:
Great catch, thanks! I'm hoping to carve out some time later this week to merge Buildalyzer PRs and do a release.

PR #155, opened 2020-12-01
Bump `Microsoft.CodeAnalysis` packages to 3.7.0

jjonescz:
In my project, I want to have:

```csproj
<PackageReference Include="Buildalyzer.Workspaces" Version="3.1.0" />
<PackageReference Include="Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation" Version="5.0.0" />
```

which gives me following version conflict error:

```txt
myproject.csproj : error NU1107: Version conflict detected for Microsoft.CodeAnalysis.Common. Install/reference Microsoft.CodeAnalysis.Common 3.7.0 directly to project myproject to resolve this issue. 
myproject.csproj : error NU1107:  myproject -> Microsoft.AspNetCore.Mvc.Razor.RuntimeCompilation 5.0.0 -> Microsoft.CodeAnalysis.Razor 5.0.0 -> Microsoft.CodeAnalysis.Common (>= 3.7.0)
myproject.csproj : error NU1107:  myproject -> Buildalyzer.Workspaces 3.1.0 -> Microsoft.CodeAnalysis.CSharp.Workspaces 3.6.0 -> Microsoft.CodeAnalysis.Common (= 3.6.0).
```

This PR fixes that issue.

### DIFF
--- a/src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
+++ b/src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
@@ -20,8 +20,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.6.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="3.6.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.7.0" />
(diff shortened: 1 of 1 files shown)


daveaglick:
Thanks! Always good to keep dependencies up to date.

FWIW I think your particular problem would also have been solved by installed `Microsoft.CodeAnalysis.Common` 3.7.0 directly in your project. The error message here is essentially saying "two of your references require different versions and we want you to be the decider by installing the correct one yourself". Thankfully in .NET Core the binding for each of the dependencies is taken care of without requiring binding redirects (usually 😬). Of course, adding the older version of `Microsoft.CodeAnalysis.Common` as the explicit reference might result in missing methods, etc. though.

In any case, happy to bump the version on my end. I'll probably take it up to 3.8.0 as well.

Issue #157, opened 2020-12-03
AddToWorkspace doesn't support source generators


svick:
Consider the following code (where `C:\code\roslyn-sdk` is a path to checked out [dotnet/roslyn-sdk](https://github.com/dotnet/roslyn-sdk/)):

```c#
var manager = new AnalyzerManager();
var analyzer = manager.GetProject(@"C:\code\roslyn-sdk\samples\CSharp\SourceGenerators\GeneratedDemo\GeneratedDemo.csproj");
AdhocWorkspace workspace = new AdhocWorkspace();
Project roslynProject = analyzer.AddToWorkspace(workspace);

foreach (var analyzerReference in roslynProject.AnalyzerReferences)
{
    Console.WriteLine(analyzerReference.FullPath);
}
```

This does not print anything; `roslynProject.AnalyzerReferences` is empty.

What I expect is that it prints references to source generator assemblies used by the project:

```
C:\Users\PetrOnderka\.nuget\packages\csvtextfieldparser\1.2.2-preview\lib\netstandard2.0\CsvTextFieldParser.dll
C:\Users\PetrOnderka\.nuget\packages\handlebars.net\1.10.1\lib\netstandard2.0\Handlebars.dll
C:\Users\PetrOnderka\.nuget\packages\newtonsoft.json\12.0.1\lib\netstandard2.0\Newtonsoft.Json.dll
C:\code\roslyn-sdk\artifacts\bin\SourceGeneratorSamples\Debug\netstandard2.0\SourceGeneratorSamples.dll
```

I plan to submit a PR that fixes this soon.

daveaglick:
Cool! Never even considered adding analyzers and generators to the workspace, but if Roslyn tracks them it definitely makes sense to add them.

PR #158, opened 2020-12-04
Added support for analyzers and source generators

svick:
Fixes https://github.com/daveaglick/Buildalyzer/issues/157.

Note that source generators use the same mechanism of "analyzer" references, so this single change fixes both.


daveaglick:
You're my hero for adding tests!

PR #160, opened 2021-01-02
Add Referenced Projects

### LINKED ISSUE #159: Automatically load ProjectReferences
Hi, I love Buildalyzer.

It would be great if I could point it at a project file and have it analyze all of the project references too.

At the moment I am running this:
```csharp
var workspace = analyzerManager.GetProject(testProj).GetWorkspace(true);
var solution = workspace.CurrentSolution;
```          
However solution only has 1 project in it. To work around this, I can scan for all `csproj` files in the repo and call:
```csharp
var projectFiles = Directory.EnumerateFiles("<path to my repo>/", "*.csproj", SearchOption.AllDirectories);

var analyzerManager = new AnalyzerManager();
foreach (var project in projectFiles)
{
    analyzerManager.GetProject(project);
}
```

I would love for this not to be required and for `GetWorkspace(true)` to automatically add the project to the analyzer.

slang25:
Fixes #159 

This change adds to the functionality when you set `addProjectReferences` to `true`. At the moment it will follow all references where you have already added the project to the solution, but if you want to create an ad-hoc solution with a csproj as the entry point, I think you'll want this behaviour to automatically add references projects to the solution.

Just to be clear, the code change code (in `GetReferencedAnalyzerProjects`) is only called when `addProjectReferences` is `true`.

### DIFF
--- a/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
+++ b/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
@@ -59,7 +59,7 @@ namespace Buildalyzer.Workspaces.Tests
         }
 
         [TestCase(false, 1)]
-        [TestCase(true, 2)]
+        [TestCase(true, 3)]
         public void AddsProjectReferences(bool addProjectReferences, int totalProjects)
         {
             // Given
--- a/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
+++ b/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
@@ -65,7 +65,6 @@ namespace Buildalyzer.Workspaces.Tests
             // Given
             StringWriter log = new StringWriter();
             IProjectAnalyzer analyzer = GetProjectAnalyzer(@"projects\LegacyFrameworkProjectWithReference\LegacyFrameworkProjectWithReference.csproj", log);
-            GetProjectAnalyzer(@"projects\LegacyFrameworkProject\LegacyFrameworkProject.csproj", log, analyzer.Manager);
 
             // When
             Workspace workspace = analyzer.GetWorkspace(addProjectReferences);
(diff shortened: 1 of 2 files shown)


tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -59,7 +59,7 @@ public void CreatesCompilationOptions()
         }
 
         [TestCase(false, 1)]
-        [TestCase(true, 2)]
+        [TestCase(true, 3)]
```
slang25, line 62: Essentially this is the change in behavior, you don't need to add child projects now when `addProjectReferences` is used. If the old behavior is still desired, perhaps we should change the parameter to an enum and support both variants?

slang25:
To counteract the guilt I'd feel bumping this PR, I bought you a coffee 😆

daveaglick:
Lol. Was already planning to take a look at the open Buildalyzer PRs this weekend, but the beverage is appreciated!

slang25:
😄 it's the least I could do.

 I think I may not need this anymore, what I need is basically this:
https://github.com/aws/codelyzer/blob/acd3c02c4b41aa0a14393fcf0ae6f4f145212f97/src/Analysis/Codelyzer.Analysis.Build/WorkspaceBuilderHelper.cs#L74-L113

but I want it to happen out of the box. When I use this PR alone, doesn't give the right results from a clean build, because I've not built the projects in the right order.

daveaglick:
Tell me more about

> When I use this PR alone, doesn't give the right results from a clean build, because I've not built the projects in the right order.

Is the idea that sometimes `analyzerResult.Manager.GetProject(x)` won't have the project yet and so we need an extra flow where if the project hasn't already been added to the workspace, _and_ the manager hasn't yet built it, we should go off and build it and then add it?

That seems totally doable, though I think you're right that we'd want to put that behind a flag (probably combined with the behavior in this PR - something like `bool buildMissingReferences`.

daveaglick:
I guess the part I'm not quite understanding is:
- `GetReferencedAnalyzerProjects()` now adds the project file to the manager (after this PR) if it wasn't already there
- Then we loop through those projects (including missing ones thanks to this PR) and call `ProjectAnalyzer.AddToWorkspace()`
- Inside `AddToWorkspace()` we call `return analyzer.Build().FirstOrDefault().AddToWorkspace(workspace, addProjectReferences);` which seems like it should do what you're thinking and build any projects that weren't already built.

So my disconnect is why that doesn't work as expected to build projects you're missing once this PR is merged.

daveaglick:
I think I get it...is the problem that we don't pick up on transitive references? I.e. if I'm adding project A to the workspace and if references project B, I'll add and build B. But if B also references C, I'll never add and build C to the workspace.

daveaglick:
I'm going to bring this discussion back to the original issue...

PR #176, opened 2021-04-13
Fixing ProjectAnalyzer exceptions when solution file is not in the source root

### LINKED ISSUE #175: ProjectAnalyzer throws when solution is not in a project parent directory
I have a big database so I don't have a common solution and instead use [SlnGen](https://github.com/microsoft/slngen) for those purposes. So I basically have temp solution files in test projects rather than one solution file in the root. 

Turns out, in this case ProjectAnalyzer throws an error in the ctor when doing some string manipulation.
```
System.ArgumentOutOfRangeException: startIndex cannot be larger than length of string. (Parameter 'startIndex')
   at System.String.Substring(Int32 startIndex, Int32 length)
   at System.String.Substring(Int32 startIndex)
   at Buildalyzer.ProjectAnalyzer..ctor(AnalyzerManager manager, String projectFilePath, ProjectInSolution projectInSolution)
   at Buildalyzer.AnalyzerManager.GetProject(String projectFilePath, ProjectInSolution projectInSolution)
   at Buildalyzer.AnalyzerManager.GetProject(String projectFilePath)
```

psfinaki:
closes #175

### DIFF
--- a/src/Buildalyzer/Construction/IProjectFile.cs
+++ b/src/Buildalyzer/Construction/IProjectFile.cs
@@ -8,6 +8,7 @@ namespace Buildalyzer.Construction
         bool IsMultiTargeted { get; }
         IReadOnlyList<IPackageReference> PackageReferences { get; }
         string Path { get; }
+        string Name { get; }
         bool RequiresNetFramework { get; }
         string[] TargetFrameworks { get; }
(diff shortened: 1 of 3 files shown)


daveaglick:
Nice fix, thanks!

PR #177, opened 2021-04-29
When SolutionFilePath is available set in Workspace Solution

markrendle:
When creating a Workspace from an AnalyzerManager created from a solution file, this change ensures that `CurrentSolution.FilePath` is set correctly.

There were some failing integration tests but I think they're due to my environment being different.

### DIFF
--- a/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
+++ b/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
@@ -29,6 +29,23 @@ namespace Buildalyzer.Workspaces.Tests
         }
 
         [Test]
+        public void LoadsSolution()
+        {
+            // Given
+            string solutionPath = GetFullPath(@"projects\TestProjects.sln");
+            StringWriter log = new StringWriter();
+            AnalyzerManager manager = new AnalyzerManager(solutionPath, new AnalyzerManagerOptions { LogWriter = log });
+
+            // When
+            Workspace workspace = manager.GetWorkspace();
+
+            // Then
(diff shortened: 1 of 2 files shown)


tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -122,18 +139,21 @@ public void SupportsAnalyzers()
...
+
             return manager.GetProject(projectPath);
         }
+
+        private static string GetFullPath(string partialPath)
+        {
+            return Path.GetFullPath(
+                Path.Combine(
+                    Path.GetDirectoryName(typeof(ProjectAnalyzerExtensionsFixture).Assembly.Location),
+                    @"..\..\..\..\" + partialPath));
```
psfinaki, line 156: Consider using `Directory.GetParent` instead, slashes are different in different operating systems.

markrendle: I didn't want to change too much, that was existing code refactored into a shared method so I assume it's been working.

daveaglick: @markrendle thanks, keeping the changes limited makes it easier for me to pull this change in without spending a lot of time I don't have right now reviewing it.

@psfinaki I think you've got a very good point though and at some point soon I'll need to do a little auditing of the code to remove slashes wherever I can. That said, I do think the `Path` family of APIs normalizes slash characters fairly well across platforms which is why this code worked well enough before.

In any case, good enough to merge and release for now :)

daveaglick:
Thanks!

PR #182, opened 2021-07-13
Fix code samples in README

eNeRGy164:
The code samples in the README are not working.

Made small tweaks so they can by copy & pasted.


psfinaki (approved):

daveaglick:
Great call, thanks! I forget when we changed to interfaces, but it was a while back and I must have totally forgotten about the docs.

Issue #185, opened 2021-08-30
Build with generated code


bernd5:
Hello,

with net 6.0 we get generated code by default (implicit global usings).
I would like to load those generated files with Buildalyzer, too.

What should I do?

slang25:
I understand that the latest previews for .NET 6 require an opt-in in the project file (certainly for preview 7), so you would need to add:
```xml
<ImplicitUsings>enable</ImplicitUsings>
```

bernd5:
In the new templates these implicit usings are enabled by default.
I created a test pr with a helloWorld sample without touching the csproj file. The test fails currently.

slang25:
I've ran the test locally here, and it fails due to a weird error, which is the same error I see on all my projects (whatever tfm):
> Could not copy the file apphost.exe

I'm not sure why, but when .NET 6 is present (i.e. installed and not suppressed by a `global.json` file) then the error crops up.
I can work around like this:
```csharp
System.Environment.SetEnvironmentVariable("UseAppHost", "false");
```

When I add that to the tests, your `TestUsings` project compiles successfully.
cc @daveaglick

slang25:
Potential fix: https://github.com/daveaglick/Buildalyzer/pull/187

phmonte:
Merge of solution accomplished

PR #186, opened 2021-08-30
Added hello-world test case for net 6.0 projects

bernd5:
See https://github.com/daveaglick/Buildalyzer/issues/185


daveaglick:
Nice - you realize this means you can never delete that repository now, right? :)

PR #187, opened 2021-09-06
Disable AppHost Creation

slang25:
In general this can be considered an optimization, but it also fixes an error that I'm running into where msbuild tries to copy the `apphost.exe`, but it doesn't exist.

I'd like to understand why it fails, but at the same time, a design-time build probably should be switching this off.


daveaglick:
Yeah, there's absolutely no reason we need to be creating a native executable for anything. 👍

PR #192, opened 2021-12-10
Add PreprocessorSymbols to analyserResult

### LINKED ISSUE #191: PreprocessorSymbolNames are missing for .NET 5+ projects
As mentioned on this page https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives I can use new symobols in preprocessor if like `NETCOREAPP3_1_OR_GREATER` which should be added if the targetframework is greater than 3.1.

If I use this in a project and build with Visual Studio the code is included. But when I analyse that same project with buildalyzer the preprocessor symbols are not added. Only the default symbols. 

## Reproduction
Reference .NET 6 and add to a file
```csharp
    public int Bit()
    {
#if NETCOREAPP3_1_OR_GREATER
        return 1 ^ 2;
#endif
    }
```
This will compile in VS and dotnet build, but not in buildalyzer

## Context

I'm not sure yet why they are missing in Buildalyzer. They should be added on SDK style projects. But there is a message saying: `For traditional, non-SDK-style projects, you have to manually configure the conditional compilation symbols for the different target frameworks in Visual Studio via the project's properties pages.`

So it might just be the case some SDK assembly is not loaded? Or if we have bad luck this is custom logic in MSBuild thats used in VS and the dotnet CLI. 

Either way I can't compile the project that builalyzer returns... I also see no workaround as the symbols are added at parse time and can't be changed once the syntax tree has been parsed.

Issue was found in https://github.com/stryker-mutator/stryker-net/issues/1828

richardwerkman:
closes #191

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -89,6 +89,13 @@ namespace Buildalyzer
                 .Select(x => x.Item2)
                 .ToArray() ?? Array.Empty<string>();
 
+        public string[] PreprocessorSymbols =>
+            _cscCommandLineArguments
+                ?.Where(x => x.Item1 == "define")
+                .SelectMany(x => x.Item2.Split(new[] { ';' }, StringSplitOptions.RemoveEmptyEntries))
+                .Select(x => x.Trim())
+                .ToArray() ?? Array.Empty<string>();
+
         public IEnumerable<string> ProjectReferences =>
             Items.TryGetValue("ProjectReference", out IProjectItem[] items)
                 ? items.Select(x => AnalyzerManager.NormalizePath(
--- a/src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
+++ b/src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
@@ -171,12 +171,7 @@ namespace Buildalyzer.Workspaces
                     CSharpParseOptions parseOptions = new CSharpParseOptions();
 
                     // Add any constants
-                    string constants = analyzerResult.GetProperty("DefineConstants");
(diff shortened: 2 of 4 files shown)


daveaglick:
Nice work! I'll try my best to get this reviewed and published later today or over the weekend.

src/Buildalyzer/AnalyzerResult.cs
```diff
@@ -89,6 +89,13 @@ internal AnalyzerResult(string projectFilePath, AnalyzerManager manager, Project
                 .Select(x => x.Item2)
                 .ToArray() ?? Array.Empty<string>();
 
+        public string[] PreprocessorSymbols =>
+            _cscCommandLineArguments
+                ?.Where(x => x.Item1 == "define")
```
psfinaki, line 94: Will it make sense to compare things ignoring the case?

psfinaki (approved):

src/Buildalyzer/AnalyzerResult.cs
richardwerkman: As far as I can see all Item1 values are lowercase. They are gathered from msbuild which doesn't really change I think. But I guess it couldn't hurt to add ignoring case. But then we should do this for all cscCommandLineArguments we use in this file

daveaglick:
Thanks, good find - I'll get a release out soon (working on it now).

richardwerkman:
Thanks for the quick merge! 😄

psfinaki:
Looking forward to the release!

daveaglick:
Hopefully forthcoming - ran into some issues while doing final integration test runs last night so I'm working through those now before pushing the new release.

psfinaki:
Thanks a lot!

richardwerkman:
Works like a charm! Thanks for the release 🎉

PR #197, opened 2022-02-13
Update Roslyn based libraries to 4.0.1 and use .NET 6 SDK with multitarget

AdaskoTheBeAsT:
Another way of doing upgrade in comparison to
https://github.com/daveaglick/Buildalyzer/pull/189

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -291,3 +291,6 @@ __pycache__/
 *.ncrunchproject
 
 cache/
+tests/repos/net5.0
+tests/repos/net6.0
+tests/repos/netcoreapp3.1
--- a/build/Buildalyzer.Build/Buildalyzer.Build.csproj
+++ b/build/Buildalyzer.Build/Buildalyzer.Build.csproj
@@ -3,6 +3,7 @@
     <PropertyGroup>
       <OutputType>Exe</OutputType>
       <TargetFramework>net6.0</TargetFramework>
+      <LangVersion>10.0</LangVersion>
     </PropertyGroup>
 
     <ItemGroup>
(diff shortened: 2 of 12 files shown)


daveaglick:
Ah, very clever using the StructuredLogger package to provide up-to-date log handling while downgrading the CodeAnalysis packages on .NET Core 3.x. I like where you’re going with this.

At this point, why even bother with the .NET 5/6 targets though? This drops .NET Framework support (which is okay) and .NET 5/6 can correctly consume a .NET Core 3.x package, so do we gain anything with the multi-targeting or **would it be sufficient just to take your changes for .NET Core 3.1 and leave it at that for now?**

I don’t like the idea of this much conditional on the target, and I have a feeling it’ll just continue to balloon as time goes on, so I don’t want to maintain it in this state. But if only targeting .NET Core 3.1 works with your approach we’ll do that and keep moving forward on that target as long as we can. If not, I’ll merge this PR for one final release before a new major release like what I was working on with a .NET 6 target.

Either way, nice work!

daveaglick:
Cool - I'm working on this right now, stripped out the .NET 5/6 stuff and focusing on your changes to support .NET Core 3.1. If that works without the 5/6 targets we'll move forward with that approach as a baseline until we really need to update to .NET 6 at some point in the future. Even though we'll still have to drop .NET Framework support (I don't really see a way around that), this at least keeps the .NET Core 3.1 support for folks who need it.

daveaglick:
Everything worked great and all tests passed, so I'm going to move forward with this change to keep us on .NET Core 3.1 for now. Very nice work!! Saved me quite a bit of trouble downstream in some other projects for now.

PR #198, opened 2022-02-13
Make AnalyzerResults deterministically ordered

0xced:
The order itself is not relevant (the `NuGetFrameworkSorter` used to sort target frameworks doesn't guarantee any specific order anyway) but the order is deterministic, meaning that several builds of the same project will always yield the same order of the target frameworks and results.

The nondeterministic nature of the current implementation made it hard to diagnose an implementation issue in Stryker.NET: https://github.com/stryker-mutator/stryker-net/issues/1899#issuecomment-1035432919

### DIFF
--- a/src/Buildalyzer/AnalyzerResults.cs
+++ b/src/Buildalyzer/AnalyzerResults.cs
@@ -1,6 +1,7 @@
 ﻿using System.Collections;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Buildalyzer
 {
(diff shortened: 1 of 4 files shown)


daveaglick:
Thanks! Looks good at first glance. Trying to get a strange test failure to clear in CI (it passes locally in Windows but not on the Windows agent, and passes the other OS agents so I'm a little perplexed) and then I'll take a closer look at this PR.

0xced:
I see that the test causing issues uses `Build(options).First()` which is non-deterministic! This is exactly what this pull request is about, so you should also try to run the tests on this PR. I think that might solve this issue. 😉

daveaglick:
FYI - I did a rebase and force push to trigger the CI build/test reporting again. Still working on getting that just right, and seemed like the only way to restart from the refreshed GitHub Actions workflows. Assuming all the tests pass I'll get this merged.

0xced:
✅ It passed! 🥳

daveaglick:
:shipit: 

I'll get the next Buildalyzer release out momentarily.

PR #213, opened 2022-09-16
Implement API to retrieve compiler command-line

### LINKED ISSUE #212: Add API to retrieve compiler command-line and arguments from `IAnalyzerResult`
Proposed API:

```csharp
interface IAnalyzerResult
{
+ /// <summary>
+ /// Gets the compiler command produced by the `CoreCompile` task.
+ /// </summary>
+ string Command { get; }
+ /// <summary>
+ /// Gets the arguments passed to the compiler produced by the `CoreCompile` task.
+ /// </summary>
+ IEnumerable<string> CompilerArguments { get; }
}
```

The use-case behind this is that I want to use Roslyn's `CSharpCommandLineParser` to get the correct compilation and parser options. Currently I have to implement my own MSBuild logger similar to https://github.com/daveaglick/Buildalyzer/blob/main/src/Buildalyzer/Logging/EventProcessor.cs to get to the raw command.

Would accept a PR that implements this feature?

siegfriedpammer:
Fixes #212

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -209,8 +209,7 @@ namespace Buildalyzer
             cmd.Arguments = new List<string>();
             cmd.ProcessedArguments = new List<(string, string)>();
 
-            bool initialCommand = true;
-            using (IEnumerator<string> enumerator = EnumerateCommandLineParts(commandLine, initialCommand, initialCommandEnd).GetEnumerator())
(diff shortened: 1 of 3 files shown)


daveaglick:
That was fast, thanks! The code looks great, so assuming the build check passes I'll merge this in an get a release out. I also need to run some integration tests locally before releasing, but that rarely fails if the CI build passes.

siegfriedpammer:
Thank you very much for merging this so fast. I just updated my project and it works like a charm.

daveaglick:
Awesome, glad to hear it!

PR #215, opened 2022-09-26
#210 A change so it can analyse Azure function apps. 

bonner-earle:
For Issue #210  
I don't fully understand how csc works, so this change is based on my observations of what is happening for the test project supplied in the defect, i.e. that the changed method is called twice both with coreCompile set to true, the first one contains the file paths for the source files. So the solution that appears to work is to aggregate the the results of the two calls together, attempted in a way to not interfere with other kinds of project, or what the app was doing before.

I have no idea which of the two commands should be retained as the command, so the last one wins.
I have run the tests and run the example project supplied on the defect with this fix and it produces an output which includes all the source files.  
On my box I still get that running a design time build of the function app throws an error, which occurs for me on the integration test that does a design time build of the test Function app, however as you don't seem to have that problem it will be interesting to see if the tests pass for you

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -184,7 +184,7 @@ namespace Buildalyzer
             ProcessedCommandLine cmd = ProcessCscCommandLine(commandLine);
 
             // Some projects can have multiple Csc calls (see #92) so if this is the one inside CoreCompile use it, otherwise use the first
-            if (coreCompile)
+            if (coreCompile || _cscCommandLineArguments == null)
             {
                 _command = cmd.Command;
                 _compilerFilePath = cmd.FileName;
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -176,16 +176,24 @@ namespace Buildalyzer
 
         internal void ProcessCscCommandLine(string commandLine, bool coreCompile)
         {
-            // Some projects can have multiple Csc calls (see #92) so if this is the one inside CoreCompile use it, otherwise use the first
-            if (string.IsNullOrWhiteSpace(commandLine) || (_cscCommandLineArguments != null && !coreCompile))
+            if (string.IsNullOrWhiteSpace(commandLine))
             {
                 return;
             }
+
             ProcessedCommandLine cmd = ProcessCscCommandLine(commandLine);
-            _command = cmd.Command;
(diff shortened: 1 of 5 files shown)


src/Buildalyzer/AnalyzerResult.cs
```diff
@@ -176,16 +176,24 @@ internal void ProcessProject(PropertiesAndItems propertiesAndItems)
...
             ProcessedCommandLine cmd = ProcessCscCommandLine(commandLine);
-            _command = cmd.Command;
-            _compilerFilePath = cmd.FileName;
-            _compilerArguments = cmd.Arguments.ToArray();
-            _cscCommandLineArguments = cmd.ProcessedArguments;
+
+            // Some projects can have multiple Csc calls (see #92) so if this is the one inside CoreCompile use it, otherwise use the first
+            if (coreCompile)
+            {
+                _command = cmd.Command;
```
siegfriedpammer, line 189: Why would you set these properties only sometimes? I would expect to get the compiler command-line of any type of project from these properties.

bonner-earle: Because I took out the condition above that if "_cscCommandLineArguments != null && !coreCompile" then the function would exit early. 
However I have a slight mismatch in the logic, so I'll adjust

when _cscCommandLineArguments != null, !coreCompile, before it exited early, now command etc not set
when _cscCommandLineArguments == null, !coreCompile,  before set command etc, now command etc not set
when _cscCommandLineArguments != null, coreCompile, before set command etc, now set command etc 
when _cscCommandLineArguments == null, coreCompile, before set command etc, now set command etc

siegfriedpammer: The point I am trying to make is that I would expect the externally visible properties `CompilerCommand`, `CompilerFilePath` and `CompilerArguments` of `IAnalyzerResult` to be in sync with the contents of the internal `_cscCommandLineArguments` list. Is there any reason why this is not possible?

bonner-earle:
After working on this and trying to get a custom version of stryker to works discovered that for function apps I only want the first call to csc.

daveaglick:
> After working on this and trying to get a custom version of stryker to work. I discovered that for function apps I only want the first call to csc.

Yeah, looks like you hit the same wall I did in #210 - there's just too much going on in the Function App extensions to MSBuild that I think it's outside the ability of Buildalyzer to analyze.

Thanks for the effort though! At least now we know.

PR #221, opened 2022-11-29
upgrade to .NET 6.0

colombod:


daveaglick:
Note that I'll need to do a little investigation before I consider this one done and am able to release another package. More specifically, there was a lot of discussion in #189 and #197 the last time we tried to bump the target. I think the situation is a little different now with another year having passed and .NET Core 3.1 going out of support, but I might need to unwind a bit of the workarounds put in place earlier.

daveaglick:
So it turns out everything from that earlier workaround had already been rolled back making this a pretty straightforward retarget. Thanks for working through the PR split @colombod - publishing this as a new major Buildalyzer version 5 to NuGet now.

PR #231, opened 2023-08-09
Support Additional files on GetWorkspace()

Corniel:
It turned out, that Additional files where not added the ProjectInfo. I created both two tests as a simple implementation. Just let me know if that is sufficient.


src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
```diff
@@ -267,17 +268,28 @@ private static CompilationOptions CreateCompilationOptions(IAnalyzerResult analy
...
+                       TextAndVersion.Create(
+                           SourceText.From(File.ReadAllText(x), Encoding.Unicode), VersionStamp.Create())),
+                   filePath: x))
             ?? Array.Empty<DocumentInfo>();
 
+        private static IEnumerable<DocumentInfo> GetAdditionalDocuments(IAnalyzerResult analyzerResult, ProjectId projectId)
+        {
+            string projectDirectory = Path.GetDirectoryName(analyzerResult.ProjectFilePath);
+
+            return GetDocuments(
```
psfinaki, line 288: You could probably reduce the number of null checks in the new code by falling back to the empty collection in this function. (just passing by)

psfinaki: Nono, I didn't mean any dramatic changes here :)

I only meant you can do:
```
var additionalFiles = analyzerResult.AdditionalFiles ?? Array.Empty<DocumentInfo>;
return GetDocuments(additionalFiles.Select...
```
...and remove all the other question marks since it's all private code.

But that's all nitpicking.

Corniel: Done!

BTW, any ETA on an updated package?

psfinaki: Cool! Well, I am not a maintainer, I am just following this repo :D 
@daveaglick should know.

Corniel: @daveaglick Any thoughts on this?

daveaglick: Soon(ish)? I really wish I had more time for OSS maintainer duties lately - my email inbox is basically just hundreds of GitHub notifications I keep in there as a queue. And then every day I run out of time 😬.

I'll try to take a look at this PR and the new one for updating MS Build dependencies tomorrow morning before heading out for the weekend. Unless there are any red flags, I'll probably just merge both and bump the major version.

Anyone on this thread have an issue with dropping < .NET 6 support in a new major? Asking here since if this change gets bundled with that one in a single new version, you won't be able to get this change without bumping the major.

Corniel: It is not my call. That being said, the [package](https://github.com/Corniel/roslyn-test-tools) I use it for, is .NET 6 (and .NET 7).

daveaglick:
Thanks for fixing this! And thanks for the review @psfinaki - that makes it easy for me to merge without too much ceremony. I'll be publishing a new minor with just this change momentarily, and then I'll publish a new major that includes the .NET 6 updates from #234 (that way this change will be available to consumers stuck on older .NET versions).

daveaglick:
The failure here is from the Cake build (the test suite currently builds a handful of other open source projects). I actually just removed that entire test fixture since keeping up with the various edge cases for full third-party codebases was too challenging. We've already got a test suite with different project types and should just continue to add to that as needed.

None of this is your problem, just noting it here for future reference :).

PR #235, opened 2023-09-04
Support nullable context options

Corniel:
I noticed that the [`NullableContextOptions`](https://sourceroslyn.io/#Microsoft.CodeAnalysis/Compilation/NullableContextOptions.cs) where not set while creating the `CSharpCompilationOptions`. With this PR, that should be fixed.


Corniel:
@daveaglick

You're loaded with work, that is pretty clear. How can I (we) help you?

Issue #228, opened 2023-05-16
Error when trying to read dotnet --info


phmonte:
I'm getting an error when reading the dotnet --info command, at certain times my cluster takes more than 4000ms to return the dotnet --info which causes Buildalyzer to fail.

Does it make sense to implement an environment variable for custom timeout values or wait for the console to respond to some value?
I can open this PR, but I would like to know if it makes sense and if there is any preference for waiting for the console or reading the timeout of some environment variable.

Just for context, I'm using dotnet stryker which has a dependency on Buildalyzer.

[Code](https://github.com/daveaglick/Buildalyzer/blob/5b732a8ce572efbba077dd08871fb609c69f94ce/src/Buildalyzer/Environment/DotnetPathResolver.cs#L63)

[Stryker Issue](https://github.com/stryker-mutator/stryker-net/issues/2488)

daveaglick:
#236 was merged recently, which includes your environment variable to extend the `dotnet --info` timeout. I also extended the default wait time in version 6.0.2 to 10 seconds (rolling out now). Can you try the latest 6.0.2 version and let me know if this issue is now resolved? Thanks!

phmonte:
I'll test it and let you know, thanks.

phmonte:
Worked perfectly.

PR #236, opened 2023-09-13
Adding timeout to the dotnet info command, using environment variables.

phmonte:
**Description**
In some scenarios 4000ms to execute the dotnet --info command is not enough.

**Objective** 
The objective is to add the possibility of passing the desired timeout via environment variable.

**Scenario**
I identified this scenario when running the solution within k8s clusters with low memory and CPU.

[Related Issue](https://github.com/daveaglick/Buildalyzer/issues/228)


daveaglick:
Finally getting around to some reviews and merges in this project, sorry for the delay! This all looks fine to me, merging now - thanks!

PR #239, opened 2023-10-10
Fix parsing for Azure Function projects

### LINKED ISSUE #210: Errors when analysing an Azure Function .net project
Hello
Running Buildalyzer on an Azure Function project fails.
An example log file. 
[logs.log](https://github.com/daveaglick/Buildalyzer/files/9221026/logs.log)
Attached is a solution containing a simple azure function project and a project which calls Buildalyzer on the azure function project used to create the example log file 
[StrykerReplication.zip](https://github.com/daveaglick/Buildalyzer/files/9220986/StrykerReplication.zip)

The example solution is called StrykerReplication as this started out life as a project to reproduce a bug in the Stryker project, see https://github.com/stryker-mutator/stryker-net/issues/1859

oviradoi:
Fixes #210 

* Remove global properties that prevent building of Azure Function projects
* Change AnalyzerResult to gather results from multiple compilation calls


tests/projects/FunctionAppIsolated/.gitignore
```diff
@@ -0,0 +1,264 @@
+## Ignore Visual Studio temporary files, build results, and
```
Corniel, line 1: Why is this file needed?

Corniel:
In general this might work. I'm under the impression that it might be worth cleaning up the build environment code first. What do you think @daveaglick ?

tests/projects/FunctionAppIsolated/.gitignore
oviradoi: The whole FunctionAppIsolated project was generated with Visual Studio, and the gitignore file as well. I don't think it's needed, and it can be removed.

tests/projects/FunctionAppIsolated/Function1.cs
```diff
@@ -0,0 +1,30 @@
+using System.Net;
+using Microsoft.Azure.Functions.Worker;
+using Microsoft.Azure.Functions.Worker.Http;
+using Microsoft.Extensions.Logging;
+
+namespace FunctionAppIsolated
+{
+    public class Function1
```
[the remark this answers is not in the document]
oviradoi: This file and the whole project was generated with Visual Studio. I'm not sure what you mean with your comment. Are you trying to say that this class should be renamed to `ExampleFunction` ?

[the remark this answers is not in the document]
Corniel: Yes. Just a bit more descriptive. `EmptyFuction` might even be better. Sorry, my remark was not that descriptive either. ;)

daveaglick:
Finally getting around to reviewing this, thanks for the hard work so far!

src/Buildalyzer/Environment/BuildEnvironment.cs
```diff
@@ -107,16 +107,12 @@ public sealed class BuildEnvironment
                 _globalProperties.Add(MsBuildProperties.DesignTimeBuild, "true");
                 _globalProperties.Add(MsBuildProperties.BuildingProject, "false"); // Supports Framework projects: https://github.com/dotnet/project-system/blob/main/docs/design-time-builds.md#determining-whether-a-target-is-running-in-a-design-time-build
                 _globalProperties.Add(MsBuildProperties.BuildProjectReferences, "false");
-                _globalProperties.Add(MsBuildProperties.SkipCompilerExecution, "true");
```
daveaglick, line 110: The changes here are really my only concern. It appears as though, in the interest of supporting Functions apps, we're essentially disabling design-time builds by turning off several of the build props that would otherwise put MSBuild into that mode. In particular, turning compiler execution back on will slow down _every_ use of Buildalyzer. Further down, turning copying to the output directory back on might risk a regression of some problems folks used to have with their built assemblies being overwritten during Buildalyzer use.

I guess in general, I'm not super comfortable changing the global props used for the default behavior since those have evolved over a long timespan. I'm wondering if there's a way to identify an Azure Functions project and make these changes, _but only_ if the project is a Functions app?

Corniel:
I advise to close this one in favor of #280

PR #242, opened 2024-01-06
Restore sorting empty target framework last

0xced:
Commit 8e85a15e324a65f9da5033f3d8a2320c76dda378 dropped the NuGet.Frameworks dependency and as a result sorting of the target frameworks behaviour changed.

Previously, the `NuGetFrameworkSorter` would sort the empty target framework last. Since 8e85a15e324a65f9da5033f3d8a2320c76dda378 the empty target framework is sorted first.

This commit restores the old behaviour of sorting the empty target framework last.

Note: Stryker.NET was inadvertently depending on this behaviour, so v6.0.1 broke Stryker.NET. I have [fixed the issue in Styker.NET](https://github.com/stryker-mutator/stryker-net/pull/2811) but I think restoring the previous sorting behaviour is a good idea anyway.

### DIFF
--- a/src/Buildalyzer/AnalyzerResults.cs
+++ b/src/Buildalyzer/AnalyzerResults.cs
@@ -25,7 +25,7 @@ namespace Buildalyzer
 
         public IAnalyzerResult this[string targetFramework] => _results[targetFramework];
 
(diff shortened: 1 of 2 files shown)


daveaglick:
Yikes, sorry! And good catch - nice work on bringing back a `TargetFrameworkComparer` that doesn't depend on NuGet.Frameworks. Merging and deploying a patch now.

PR #244, opened 2024-01-28
Bug: Additional files not regonized

Corniel:
As mentioned in #243, I noticed that in some projects additional files (and other properties as well I guess, but that was not what I was investigating) are not recognized.

I've added created this PR, that adds a replay where the additional files are not available in the `AnalyzerResult`. 

1. I noticed (so far) that `EventProcessor.MessageRaised(object sender, BuildMessageEventArgs e)` is triggered only for v6.0.
2. For the (existing) test that succeeds, only one TFM is detected, for this failing test, both v6.0 and v5.0 are added. The latter lacks the additional files, but is used to build anyway.
3. If change the targets to only build .NET 5.0, or for .NET 8.0 and .NET 6.0 combined, everything works just fine.

I'll update mine findings here.

For some background, I use Buildalyzer to test Roslyn Analyzers that check for issues in .NET project files, such as MS Build project files, and RESX files. Access to the additional files is the perquisite for those tests to work. See: [dotnet-project-file-analyzers](https://github.com/dotnet-project-file-analyzers/dotnet-project-file-analyzers/pull/99)

### DIFF
--- a/Buildalyzer.sln
+++ b/Buildalyzer.sln
@@ -41,6 +41,11 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "workflows", "workflows", "{
 		.github\workflows\test-report.yml = .github\workflows\test-report.yml
 	EndProjectSection
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "projects", "projects", "{A7A6E392-4132-4773-97DC-1815D773CFA5}"
+	ProjectSection(SolutionItems) = preProject
+		tests\projects\ProjectFileAsAdditionalFile\ProjectFileAsAdditionalFile.csproj = tests\projects\ProjectFileAsAdditionalFile\ProjectFileAsAdditionalFile.csproj
+	EndProjectSection
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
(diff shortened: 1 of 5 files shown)


Corniel:
@phmonte with a nice `[DebuggerDisplay]` attribute, I was able to finally find what is going wrong: for the .NET 5.0 target, nothing is compiled: No source files, no additional files. Would you be so kind to also take a look?

phmonte:
I think I found the problem, you can see it in my commit.
But, some tests broke and I still don't think it's the best solution.

Corniel:
I've been working on a new way of tracing/logging, that also is really in draft mode, but I think it looks promising. I've created another (draft) pr for that: #286

phmonte:
Perfect, I made a small change to the test project, keeping only 1 TargetFrameworks field and removing the TargetFramework, looking at the documentation [here ](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props).

I believe the additional files worked and the test passed.

I noticed that the MessageRaised event is called according to the number of TargetFramework(s), when placing the 2 fields, it considered only the first one, as it says in the documentation

Corniel:
@phmonte It defined wrong on purpose. ;) It is part of a test for [.NET project file analyzers](https://dotnet-project-file-analyzers.github.io). I was suspecting that it might not work because of this. Sorry, I should have specified that.

PR #245, opened 2024-01-28
Proposal: represent  of CompilerItems and CompilerProperties as dedicated types

Corniel:
While debugging #244, I noticed that debugging which compiler properties where active was not really clear. So similar to #243, I introduced two dedicated (read-only) types that contain the properties and the items. Potentiality, it can replace the dictionaries in the `AnalyzerResult`. That is, obviously a breaking change.

Under the hood, `CompilerProperties` and `CompilerItemsCollection` are dictionaries (with case-insensitive keys), but as I think (that can be challenged) having the values without the related key have less value in this case, the `TryGet()` method returns a `KeyValuePair` like struct that combines the two.

Let me know what you think.

### DIFF
--- a/src/Buildalyzer/AnalyzerResult.cs
+++ b/src/Buildalyzer/AnalyzerResult.cs
@@ -1,12 +1,6 @@
-using System;
-using System.Collections;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using System.Text;
 using Buildalyzer.Construction;
 using Buildalyzer.Logging;
-using Microsoft.Build.Framework;
 
 namespace Buildalyzer;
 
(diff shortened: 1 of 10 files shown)


daveaglick:
As with #243, I love where your thinking with these changes but just don't have the bandwidth to bring them in. https://github.com/daveaglick/Buildalyzer/issues/247 is a request for a new maintainer(s) and I think it makes sense to let whoever takes over the project decide the direction of open PRs. Thanks for the work, and hopefully we can find a maintainer soon and get this merged.

Corniel:
@phmonte I think this one is also ready for review. This one should be easy (I think(.

phmonte (approved):

Issue #249, opened 2024-02-16
Coding style, static code analysis, and versioning


Corniel:
## Framework version
The current target framework is .NET 6. Its support will be dropped end of this year. I would suggest to add .NET 8 asap, and drop .NET 6 support at the moment a breaking change has to be made.

## Language version
The current language version is 10. I hope we can upgrade to C# 12. There are so many nice features that could be benefited from.

## Warning as error
In the build properties, `WarningAsError` has been set to `true`. Although I understand the reasoning of that setting, I strongly believe that is counter productive. You want to be notified that you have a `TODO`, but it should not make you build fail. Without changing the  `WarningAsError` locally, I can not even build the current solution, as my IDE (Visual Studio) triggers some external analyzers (including a spelling checker) that result in build errors.

## Static code analysis
I would love to add:
* [Sonar Analyzers](https://www.nuget.org/packages/SonarAnalyzer.CSharp) (you can always have a debate on individual rules).
* [QW0003: Decorate pure functions](https://github.com/Qowaiv/qowaiv-analyzers/blob/main/rules/QW0003.md)
* [QW0005: Seal concrete classes unless designed for inheritance](https://github.com/Qowaiv/qowaiv-analyzers/blob/main/rules/QW0005.md)
* [QW0007 - Use file-scoped namespace declarations](https://github.com/Qowaiv/qowaiv-analyzers/blob/main/rules/QW0007.md)

## Editorconfig
The readability of the editor config can improved by adding a comment about the rule **behind** its definition. Most people - me included - do not now the identifiers of those analyzers by hart.
```
dotnet_diagnostic.S3218.severity = warning # Inner class members should not shadow outer class "static" or type members
dotnet_diagnostic.S3257.severity = warning # Declarations and initializations should be as concise as possible
dotnet_diagnostic.S3776.severity = warning # Cognitive Complexity of methods should not be too high
dotnet_diagnostic.S6354.severity = warning # Use a testable (date) time provider instead
```

As a side note, I personally really dislike RCS1008, as using `var` helps keeping code short and aligned. Can this be reconsidered?

Long story short, I love this project, and would like to keep participating. I think that this suggestions could make working with it even better.

phmonte:
Hi @Corniel, thank you very much for contributing, I’m very happy!

### Framework version and Language version
.NET 8.0 and language 12.0
I have no doubt that we really needed to do this update asap, I'm already looking at your PR.

### WarningAsError
[...]

PR #250, opened 2024-02-16
Update to .NET 8 and C# 12

Corniel:
As suggested in #249 , I took the liberty to update the target frameworks to .NET 8 (and for the published packages, .NET 6 is the secondary target), and use C# 12.

While at it, I updated the analyzer and Unit Test dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
           include-prerelease: true
           dotnet-version: |
             3.1.x
-            7.0.x
+            8.0.x
       - name: Build and Test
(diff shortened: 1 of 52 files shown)


Corniel:
@phmonte I also converted al `namespace` declarations to file scoped ones.

phmonte (approved):

Corniel:
@phmonte I would expected you to have squach merged the branch, but you did not. Was that on purpose?

phmonte:
The repository had squash disabled, I'll enable it, sorry!
I can do the squash manually.

phmonte:
Done

PR #251, opened 2024-02-18
Use NuGetFramework to deal with TargetFramework logic

Corniel:
`NuGetFramework` represents the target framework and its domain concerns. Thanks for the suggestion @rouke-broersma !

### DIFF
--- a/src/Buildalyzer/Buildalyzer.csproj
+++ b/src/Buildalyzer/Buildalyzer.csproj
@@ -12,6 +12,7 @@
     <PackageReference Include="Microsoft.Build" Version="17.0.1" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.0.1" />
     <PackageReference Include="MSBuild.StructuredLogger" Version="2.1.815" Aliases="StructuredLogger" />
+    <PackageReference Include="NuGet.Frameworks" Version="6.9.1" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


rouke-broersma:
Note that https://www.nuget.org/packages/NuGet.Frameworks (https://github.com/NuGet/NuGet.Client/blob/dev/src/NuGet.Core/NuGet.Frameworks/NuGetFramework.cs) might also work for this usecase, then you won't have to maintain the parsing yourself.

rouke-broersma:
@Corniel @phmonte I understand the concern however note that it is not a dependency on Nuget, it is only the TargetFramework parsing and comparison as used by Nuget. It's a separate package that as far as I can tell does not contain anything other than functionality to work with TargetFramework, such as parsing strings to types and implementing IComparer.

See: https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Core/NuGet.Frameworks

The benefit would be that you would use the same resources to work with TargetFramework as Nuget does, so it's more or less guaranteed to be correct.

It was just a suggestion, I have no hard feelings one way or the other. If you prefer to keep your own heuristics that's fine of course.

Corniel:
@rouke-broersma My personal goal is to structure this code a bit. It does awesome stuff, but is has become a challenge to maintain. Using [NuGet.Frameworks](https://github.com/NuGet/NuGet.Client/tree/dev/src/NuGet.Core/NuGet.Frameworks) might be a good step. So thanks again for the suggestion.

Corniel:
@phmonte Yep. Eventually, where the targetFramework is used as a string I would like to replace that with this class too, but as mentioned before: baby steps is the best approach here (and I will lead to breaking changes).

phmonte (approved):

PR #253, opened 2024-02-20
Drop `EmptyDisposable`

Corniel:
Providing `EmptyDisposable` as part of the public API should not be part of Buildalyzer. I've decorated it as `[Obsolete]` and created a private alternative for the place where it was actually used.

### DIFF
--- a/src/Buildalyzer/EmptyDisposable.cs
+++ b/src/Buildalyzer/EmptyDisposable.cs
@@ -2,6 +2,7 @@
 
 namespace Buildalyzer;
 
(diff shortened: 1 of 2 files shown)


phmonte:
Hi @Corniel,
After merging some PRs we had conflicts, could you resolve them?
I believe I can resolve them manually, but if I can resolve them, I think it's safer.

Corniel:
@phmonte Done.

phmonte (approved):

phmonte (approved):

PR #257, opened 2024-02-27
Define a lightweight struct to represent paths

Corniel:
Buildalyzer both has to work on UNIX and Windows bases file systems. To deal with (potential) case sensitivity of paths, It also deals with different directory separator paths. The internal representation of `IOPath` uses UNIX's `'/'`, and the `.ToString()` representation is file system based.

My plan is to start using this struct instead of `string`'s for properties that represent IO Paths. Personally, I'm also a big fan of `System.IO.FileInfo`, and `System.IO.DirectoryInfo`, hence the methods to convert into one of these. The reason not use them directly has to do with the fact that both types handle equality based on their reference, not the *thing* the represent on the file system.


phmonte (approved):

Corniel:
@phmonte Ready for merge? (I'd like to use this in #243 )

phmonte:
@Corniel done, thanks!

PR #258, opened 2024-03-09
Drop `BuildEventArgsReaderProxy`

Corniel:
It is internal and seems not to be used at all. Unless, I'm mistaken, this is dead code.


Corniel:
@daveaglick Can you confirm this, or explain why it still has purpose?

Corniel:
@daveaglick If you also can not think of any reason to keep this code, let just remove it.

Corniel:
@phmonte I think it is safe to merge this one too.

PR #260, opened 2024-03-09
Create a test context to take away a lot of the repeating plumbing code

Corniel:
Investigating the unit tests to test the Analyzers, a lot of repeating plumbing code is used. To reduce this, I created a test context. The idea is that is should take care of (most) of the plumbing code:

``` C#
[Test]
public void Some_test()
{
    using var ctx = Context.ForProject("MyProject.csproj");

    ctx.Analyzer.Build().First().AdditionalFiles
        .Should().BeEquivalentTo("message.txt");
}
```

The contexts ensures that previous created artifacts (in the `bin`, and the `obj` folder) are removed, and logs the Logger to the console if run in `DEBUG` mode. This will (still) prevent the build logs from being flooded, but gives a nice developer experience while running the tests. What do you think of this?

Obviously, naming is a thing here, and I'm open to suggestions for improvement.


phmonte:
@Corniel, I liked it, I don't like many projects within the solution, but it is necessary to share with `Buildalyzer.Workspaces.Tests`.

Is there any reason to keep the old GetProjectAnalyzer method?

I liked the change in the name of the tests too.

Corniel:
> @Corniel, I liked it, I don't like many projects within the solution, but it is necessary to share with `Buildalyzer.Workspaces.Tests`.

Indeed, otherwise I would not have created a separate project.

> Is there any reason to keep the old GetProjectAnalyzer method?

I would say no. But I did not want to change everthing before we agreed on the way to continue.

Corniel:
Ready to merge. No extra project anymore.

PR #262, opened 2024-03-10
Drop `InterlockedBool`

Corniel:
It is internal and seems not to be used at all. Unless, I'm mistaken, this is dead code. Just as `BuildEventArgsReaderProxy`. See #258


Corniel:
@phmonte Can we drop this unused code?

phmonte:
@Corniel Yes, can I do the merge or are there still any pending commits?

Corniel:
This is all.

PR #263, opened 2024-04-07
feat: adding cache to dotnet --info content

phmonte:
### Feature: 
Cache dotnet --info result per project

### Description
In some scenarios, executing dotnet --info may take longer than desired, especially when executed within containers with few resources.

This pr adds a cache where the key is the full path of the project.

### DIFF
--- a/src/Buildalyzer/Caching/DotnetInfoCache.cs
+++ b/src/Buildalyzer/Caching/DotnetInfoCache.cs
@@ -6,6 +6,7 @@ namespace Buildalyzer.Caching
     {
         private static readonly Dictionary<string, IReadOnlyCollection<string>> Cache = new Dictionary<string, IReadOnlyCollection<string>>();
 
+        [Pure]
         public static IReadOnlyCollection<string> GetCache(string path)
         {
             return Cache.TryGetValue(path, out IReadOnlyCollection<string> values) ? values : null;
--- a/src/Buildalyzer/Caching/DotnetInfoCache.cs
+++ b/src/Buildalyzer/Caching/DotnetInfoCache.cs
@@ -1,0 +1,20 @@
+﻿namespace Buildalyzer.Caching
+{
+    internal static class DotnetInfoCache
+    {
+        private static Dictionary<string, List<string>> cache = new Dictionary<string, List<string>>();
+
+        [Pure]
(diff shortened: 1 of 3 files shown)


src/Buildalyzer/Caching/DotnetInfoCache.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace Buildalyzer.Caching
+{
+    internal static class DotnetInfoCache
+    {
+        private static Dictionary<string, List<string>> cache = new Dictionary<string, List<string>>();
+
+        [Pure]
+        public static List<string> GetCache(string path)
+        {
+            return cache.FirstOrDefault(x => x.Key == path).Value;
```
Corniel, line 10: On null this will fail. And we have a dictionary, so please use cache.TryGet()

src/Buildalyzer/Caching/DotnetInfoCache.cs
```diff
@@ -0,0 +1,20 @@
...
+        }
+
+        [Pure]
```
Corniel, line 13: This is not a pure method.

phmonte: @Corniel, I really believe that neither of the 2 would be Pure, correct? as a requirement would only be local variables.

Corniel: The question is always: If I do not use the result of the call, is the call still useful? Of not, the method is pure.

phmonte:
@Corniel do you have any other considerations?

Corniel:
After further inspection, I wondered: why caching the lines? If the lines don't change, the Info does not change either. As the DotNetInfo is immutable, I would argue that the result of the parsing should be cached, not the input.

phmonte:
@Corniel thanks, it would really make sense to cache the output, I'll make the change.

src/Buildalyzer/Caching/DotnetInfoCache.cs
```diff
@@ -0,0 +1,30 @@
+﻿using Microsoft.Build.Framework;
+
+namespace Buildalyzer.Caching
```
Corniel, line 3: Don't forget to use a files coped namespace.

PR #267, opened 2024-04-15
Implement caching in .NET info resolver

Corniel:
I thought that #263 could be implemented like this. I'm not sure yet what test to add for this.


phmonte:
Thanks for the contribution, About the tests, I think of something just to ensure that the data from sample1.csproj is not being sent to queries from sample2.csproj, what do you think?

Corniel:
@phmonte I was not sure, hence my question.

PR #268, opened 2024-04-15
Use null-logger

Corniel:
Using `NullLogger` leads to cleaner code than using the `?` condtional operator.


Corniel:
@phmonte A small reminder.

PR #269, opened 2024-04-17
fix: adding alternative for publish single file

phmonte:


slang25:
Another approach might be to embed the logger dlls into Buildalyzer and have it extract them to a temporary location. It does add moving parts though.

PR #272, opened 2024-05-07
Fix hang if build does not start

### LINKED ISSUE #271: Buildalyzer hangs if build does not start
@martincostello found that adding global properties with `;` in would result in Buildalyzer hanging.

I've done some investigating and have created a minimal repro here:
https://github.com/phmonte/Buildalyzer/compare/main...slang25:Buildalyzer:msbuild-hang-repro

If you run that test you will see that it just hangs. There are really two issues here, the first is minor, which is that we should escape [the property values as outlined here](https://github.com/dotnet/msbuild/issues/471#issuecomment-366268743) to work around the fact that msbuild doesn't like `;` in property values.

The second is a bit more fundamental, if the build doesn't start for _whatever_ reason, Buildalyzer should not hang. I've been thinking about this and looking into the best way of supporting this, and there's not an obvious fix due to the way `AnonymousPipeLoggerServer` works, but the simple workaround I've found is to dispose of the underlying stream when:
1. We've received no build events
2. The process has exited with a non-zero return code

With this in place we should still get an overall result from Buildalyzer, but with an `OverallSuccess` of `false` as expected.

I'll create two PRs, one fixing the value escaping, the other with a (not ideal) fix for the general hang scenario.

slang25:
Fixes #271

There are other solutions to this, like running `ReadAll` in a `Thread` or `Task` and aborting that. I've also looked at the cancellation token and using that, however it's not being used effectively in `ReadAll` and I don't think there is simple solution as `AnonymousPipeLoggerServer` doesn't support cooperative cancellation as far as I can see.

There may be a memory leak here with the the two events being subscribed to, not sure if we want to try and engineer a better solution.

### DIFF
--- a/src/Buildalyzer/ProjectAnalyzer.cs
+++ b/src/Buildalyzer/ProjectAnalyzer.cs
@@ -153,6 +153,8 @@ public class ProjectAnalyzer : IProjectAnalyzer
         using (CancellationTokenSource cancellation = new CancellationTokenSource())
         {
             using var pipeLogger = new AnonymousPipeLoggerServer(cancellation.Token);
+            bool receivedAnyEvent = false;
+            pipeLogger.AnyEventRaised += (_, _) => receivedAnyEvent = true;
             using var eventProcessor = new EventProcessor(Manager, this, BuildLoggers, pipeLogger, results != null);
 
             // Run MSBuild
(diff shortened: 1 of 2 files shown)


src/Buildalyzer/ProjectAnalyzer.cs
```diff
@@ -170,8 +172,22 @@ public IAnalyzerResults Build(string[] targetFrameworks, BuildEnvironment buildE
                 GetEffectiveEnvironmentVariables(buildEnvironment),
                 Manager.LoggerFactory))
             {
+                processRunner.Exited += () =>
+                {
```
Corniel, line 176: I'm not sure if there is a convention specified for this, but I prefer a method being assigned here.

slang25: I've changed these event handlers to local functions 👍

phmonte:
Thanks for the contribution @slang25 .

PR #273, opened 2024-05-07
Escape semicolon in property values

### LINKED ISSUE #271: Buildalyzer hangs if build does not start

slang25:
Fixes #271

### DIFF
--- a/src/Buildalyzer/ProjectAnalyzer.cs
+++ b/src/Buildalyzer/ProjectAnalyzer.cs
@@ -250,7 +250,7 @@ public class ProjectAnalyzer : IProjectAnalyzer
         {
             argumentsList.Add(
                 propertyArgStart
-                    + $":{string.Join(";", effectiveGlobalProperties.Select(x => $"{x.Key}={FormatArgument(x.Value)}"))}");
+                    + $":{string.Join(";", effectiveGlobalProperties.Select(x => $"{x.Key}={FormatArgument(x.Value).Replace(";", "%3B")}"))}");
         }
 
         // Get the logger arguments (/l)


phmonte:
@slang25  do you see a way to create a unit test for this scenario?

I believe this problem will not only happen with the ";"

phmonte (approved):

slang25:
I'll follow up with some tests and a fuller PR soon 🙂

softlion:
> @slang25 do you see a way to create a unit test for this scenario?
> 
> I believe this problem will not only happen with the ";"

This happens because of how linux sh/bash interpreter works.

You only need to escape ";" "," and space.

PR #275, opened 2024-05-25
chore: removing some warnings

phmonte:
## Description
The objective is to clean up the project by resolving most of the warnings, it's a long job, I'm trying to solve it in parts, I've been trying to do it in a single PR for a few days, but apparently it won't be possible, we have too many.

Next steps:
Evaluate other warnings.

### DIFF
--- a/src/Buildalyzer/Logging/EventProcessor.cs
+++ b/src/Buildalyzer/Logging/EventProcessor.cs
@@ -1,8 +1,4 @@
 extern alias StructuredLogger;
-
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using Microsoft.Build.Framework;
 using Microsoft.Extensions.Logging;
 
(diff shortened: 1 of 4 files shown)


src/Buildalyzer/Logging/EventProcessor.cs
```diff
@@ -100,7 +96,7 @@ private void ProjectStarted(object sender, ProjectStartedEventArgs e)
 
             // Get the TFM for this project
             // use an empty string if no target framework was found, for example in case of C++ projects with VS >= 2022
-            var tfm = propertiesAndItems?.Properties.TryGet("TargetFrameworkMoniker")?.StringValue
+            string tfm = propertiesAndItems?.Properties.TryGet("TargetFrameworkMoniker")?.StringValue
```
Corniel, line 99: I thought `var` was okay now? Or is it this particular line where you want `string` instead?

phmonte: var is ok, I removed the warning in the config editor too.

Issue #278, opened 2024-07-12
Making the library language-agnostic


LPeter1997:
Hey, this looks like a lovely library, than you for easing the pain of setting up design-time builds!

We are working on a new .NET language, [Draco](https://github.com/Draco-lang/Compiler) and we'd love to use this library to extract references for example for our language server. Unfortunately from what I gathered, the library has only the 3 main .NET langs hardcoded for command parsing: C#, VB and F#, so we can't exactly extract that data for a custom language.

Is there any chance for this limitation to be eliminated? While I didn't dig deep into the library, I imagine we'd need to introduce an interface like so:
```cs
interface ICommandParser
{
    // Lang name tag, optional
    string Language { get; }

    // Returns true, if this parser can handle the compiler that's being invoked. Replaces ProcessFscCommandLine, ProcessCscCommandLine, ...
    bool IsLanguageCompilerInvocation(object sender, BuildMessageEventArgs e);

    // Replaces Compiler.CommandLine.Parse
    CompilerCommand Parse(DirectoryInfo? baseDir, string commandLine);
}
```

And then have a collection of these language parsers, `AnalyzerManager` or `AnalyzerManagerOptions` looks like an OK place for this. By default, the existing languages (C#, VB and F#) would have an implementation for this interface, but consumers of the package could implement this interface for any language they'd want.

Let me know what you think!

phmonte:
Hello @LPeter1997 , I don't see any problems, this limitation exists because we use msbuild, as you said, it is necessary to implement the equivalent methods for other languages.

phmonte:
@LPeter1997 have you seen Corniel's implementation? Do you think it makes sense to you?

PR #279, opened 2024-07-12
Making the library language/compiler agnostic

LPeter1997:
## Description
As I have raised in #278 , it would be nice to make the library extensible and language-agnostic, since it's pretty easy to develop .NET languages. Long story short, as I've started fleshing out the idea, basically implemented my proposed solution. Users can register `ICompilerOptionsParser`s through `AnalyzerManagerOptions` to extend with a compiler. The 3 compilers already supported (Roslyns C# and VB, FSC) are in the list by default.

### DIFF
--- a/src/Buildalyzer/Compiler/CscOptionsParser.cs
+++ b/src/Buildalyzer/Compiler/CscOptionsParser.cs
@@ -43,6 +43,9 @@ public sealed class CscOptionsParser : ICompilerOptionsParser
         var command = new CSharpCompilerCommand()
         {
             CommandLineArguments = arguments,
+            Text = commandLine,
+            CompilerLocation = location,
+            Arguments = args.ToImmutableArray(),
         };
         return RoslynParser.Enrich(command, arguments);
     }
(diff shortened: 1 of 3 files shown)


src/Buildalyzer/Compiler/CompilerLanguage.cs
```diff
@@ -1,17 +0,0 @@
-﻿namespace Buildalyzer;
-
-/// <summary>The compiler language.</summary>
```
Corniel, line 3: I disagree with dropping this. I would propose to keep `CompilerLanguage` but convert it to a (Single Value Object) readonly struct. That was not needed yet, but with your proposed solution, it will be.

Corniel:
I really like the idea of opening up Buildalizer for different languages. I'm not sure yet about the proposed solution. There are a lot of changes in here, all requiring attention. It ols means a lot of breaking changes. We should investigate how we reduce the pain for users when they have to migrate.

phmonte:
@LPeter1997  It's a big change, as I mentioned in the issue, I like the idea, but we need to evaluate this PR very carefully to ensure the next steps.
This is on my list, I just ask for patience, I promise to evaluate it as quickly as possible.

Corniel:
I tried an alternative approach, that involves way less drastic changes. See #285

Corniel:
Can you check if you can work with the API proposed in #306 ?

PR #280, opened 2024-07-16
Add support for azure function

dupdob:
## Description
This PR adds support for AzureFunctionProject

A simple integration test project is provided as well.

The logic change is: buildalyzer now keeps the first compilation step, disregarding its origin or the last compilation step when triggred by the main project file.
AzureFunction project triggers a compilation step from a temporary subproject to generate a dummy assembly (which purposes has not been identified) and which should not be reported.


src/Buildalyzer/Logging/EventProcessor.cs
```diff
@@ -146,7 +146,8 @@ private void TargetFinished(object sender, TargetFinishedEventArgs e)
     private void MessageRaised(object sender, BuildMessageEventArgs e)
     {
         AnalyzerResult result = _currentResult.Count == 0 ? null : _currentResult.Peek();
-        if (result is object)
+        // do not process any sub project
+        if (result is object && (string.IsNullOrEmpty(result.Command) || AnalyzerManager.NormalizePath(e.ProjectFile) == _projectFilePath))
```
Corniel, line 150: Preferably, define a local function that does this, so that the comment is not needed due to the self explanatory name.

Corniel (changes requested):
I think the example project would benefit from some more descriptive/representative naming.

src/Buildalyzer/Logging/EventProcessor.cs
dupdob: done
Create a `IsRelevant' local function + method refactor to increase readibility

dupdob:
> I think the example project would benefit from some more descriptive/representative naming.
like what?

phmonte:
@dupdob Thank you for contributing, we are evaluating the PR, I saw that it is related to a Stryker issue, we will try to finalize it as quickly as possible.

tests/projects/AzureFunctionProject/Function1.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
+
+namespace StrykerReplication
```
Corniel, line 5: Use global namespace;

Corniel: Allign the namespace with the project file, unless there is a reason not to so. (And that might then require an explanation).

tests/projects/AzureFunctionProject/Function1.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public class TestFunction
+    {
+        [FunctionName("Function1")]
```
Corniel, line 9:
```suggestion
        [FunctionName(nameof(TestFunction))]
```

tests/projects/AzureFunctionProject/Program.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Microsoft.Extensions.Hosting;
+
+namespace StrykerReplication
```
Corniel, line 3:
```suggestion
namespace AzureFunctionProject;
```

Corniel (changes requested):
Please update the example project.

tests/projects/AzureFunctionProject/Function1.cs
```diff
@@ -0,0 +1,23 @@
+using Microsoft.Azure.WebJobs;
+using Microsoft.Extensions.Logging;
+using System.Threading.Tasks;
+
+namespace StrykerReplication
+{
+    public class TestFunction
```
Corniel, line 7: The filename should be changed accordingly to the class name.

tests/projects/AzureFunctionProject/Function1.cs
dupdob: done, thanks for the feedback

tests/projects/AzureFunctionProject/Function1.cs
dupdob: done

Corniel (approved):
LGTM!

PR #282, opened 2024-07-24
Updates MSBuild references

### LINKED ISSUE #281: Buildalyzer should be able to build projects with C# 12 language features
I'm trying to build a .net 8 project which uses some of the new C# 12 language features such as [collection expressions](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-12#collection-expressions), the build fails and the usage of the new features are reported as comipile errors.

The following does not compile
```csharp
    internal class Program
    {
        static void Main(string[] args)
        {
            List<string> list = [];
        }
    }
```

I directly referneced the following dependencies and the code now builds
```xml
    <PackageReference Include="Microsoft.Build" Version="17.10.4" />
    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.10.4" />
    <PackageReference Include="Microsoft.CodeAnalysis.Common" Version="4.10.0" />
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="4.10.0" />
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="4.10.0" />
    <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.10.0" />
```

The required dependencies should be updated to support C#12 language features

REscobar:
Fixes #281

## Description
This PR updates MSbuild references to support C# 12 features which are not buildable using latest public version of Buildalyzer

A test case was added which would fail withouth this commit

### DIFF
--- a/src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
+++ b/src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
@@ -19,8 +19,8 @@ ToBeReleased
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.4.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="4.4.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="4.10.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Buildalyzer/Buildalyzer.csproj
+++ b/src/Buildalyzer/Buildalyzer.csproj
@@ -12,8 +12,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Build" Version="17.0.1" />
-    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.0.1" />
+    <PackageReference Include="Microsoft.Build" Version="17.10.4" />
+    <PackageReference Include="Microsoft.Build.Tasks.Core" Version="17.10.4" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="[4,)" />
     <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic" Version="[4,)" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="6.0.0" />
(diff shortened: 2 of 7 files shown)


Corniel:
This does not work when you just update these dependencies yourself in your own project?

REscobar:
Setting up a project, referencig Buildalyzer and updating those references locally does allow Buildalyzer to build projects with C# 12 features.

But I'm not using Buildalyzer directly, I'm using https://github.com/eNeRGy164/LivingDocumentation, which is a dotnet command line tool, to analyze my projects but I'm unable to do so beacuse the referenced and the latest version of Buildalyzer can't build them.

phmonte:
@REscobar  Thank you very much for contributing, I intend to evaluate your PR this week, I need to understand if we really won't have a backwards compatibility problem, there are some situations that are not possible to simulate in unit tests.

@Corniel do you see any problem? I appreciate your opinion.

Corniel:
If we take a look at: [Roslyn's release notes](https://github.com/dotnet/roslyn/blob/main/docs/wiki/NuGet-packages.md), we can conclude that v`4.4.0` (our current version) is almost 2 years old, which supports .NET 7.0 and C# 11. Although I would argue by explicitly including v`4.10.0` it should work for .NET 8.0 and C# 12 too, sticking to a version that supports up to a framework version that Microsoft does not support anymore, makes no sense.

We have no paying customers (as far as I know) that would have issues with old systems relying on this older version that can not update. So, I think we should do it.

That being said: I think we should define a strategy on how to deal with this dependencies: Should we always aim to go for the latest stable version out there?

src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
```diff
@@ -19,8 +19,8 @@ ToBeReleased
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.4.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="4.4.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.10.0" />
```
Corniel, line 22: Did you try this one (`[4,)`) too? Did this work as well?

REscobar: It does not work with `[4,)`, version `4.0.1` gets installed and fails to build a project with C#12 features

Corniel (approved):
I'm in favour of merging this.

phmonte (approved):

PR #288, opened 2024-09-08
Clean-up code by introducing String compare helpers

Corniel:
To improve on the readability when comparing strings case insensitive, I've created 3 extension methods:

``` C#
namespace System;

internal static class BuildalyzerStringExtensions
{
    public static bool IsMatch(this string? self, string? value);
    public static bool IsMatchStart(this string self, string value);
    public static bool IsMatchEnd(this string self, string value);
}
```

For obvious reasons this helpers are kept internal, as this should not be a deliverable of Buildalyzer.


Corniel:
@phmonte Can we agree on the usefulness of this?

phmonte (approved):

PR #292, opened 2024-11-01
Output type is ignored

Corniel:
While working on something else I got this issue the other way around. Anyway, when a build project with `<OutputType>exe</OutputType>` when no entry point is defined, I would expect the analysis to fail:

> CSC : error CS5001: Program does not contain a static 'Main' method suitable for an entry point \\buildalyzer\tests\projects\OutputTypeExe\OutputTypeExe.csproj

But that is not the case. This PR is first a reproduction. I'm now looking into why this is happening.


Corniel:
It turns out that I was just checking things in design-time mode. I've altered the test, and keep it in, as it is a useful test. But it was not a bug, but a not well understood (from my sight) feature.

PR #294, opened 2024-11-12
Add explicit support for assembly aliases

dupdob:
## Description
This PR adds explicit support for [assembly aliases](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/inputs#references), that is the ability to add a prefix (_an alias_) to the namespace(s) of one or more assemblies in order to deal with name conflict.

This PR adds a `ReferenceAliases` property to `IAnalyzerResult` interface. This property exposes a dictionary providing the list of aliases for every aliased references. If a referenced assembly is not aliased, it will not be listed in this dictionary.

This property is also used for Roslyn Workspace building.

## Tests
A dedicated test project (`SdkNet8Alias`) has been added and is used in integration tests and workspace tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Build and Test
         run: dotnet test --logger "trx;LogFileName=test-results.trx"
       - name: Upload Test Results
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: success() || failure()
         with:
           name: test-results-${{ matrix.os }}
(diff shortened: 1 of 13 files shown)


Corniel (approved):
LGTM

phmonte:
@dupdob Once we get the draft out, we can do the merge

dupdob:
Thanks
I put it in draft because one of the test was failing. I wanted to include an alias project reference, but it turns out it does not work for integration tests: the dependent assemblies are not built, hence not available in the workspace.
So I removed it, and everything is green again.

rouke-broersma:
@phmonte think you could find some time to release this in the near future? We could use this feature in stryker

phmonte:
@rouke-broersma yes, I can generate a new version today.

phmonte:
@rouke-broersma @Corniel and @dupdob

I'm having a compatibility issue between versions (which was expected). I'm analyzing it carefully so that we don't have any surprises.

I plan to publish the new version this weekend.

phmonte:
New version published [7.1.0]( https://www.nuget.org/packages/Buildalyzer/7.1.0), thanks.

PR #300, opened 2025-02-06
Use MacOS-13

Corniel:
The builds now fail stating that MacOS-12 has been deprecated.


Corniel:
@phmonte I would like to see this merged as quick as possible.

phmonte (approved):

PR #305, opened 2025-06-30
Handle version child element in PackageReference parsing

cslong:
## Description
This PR updates the PackageReference constructor to include the version if found in a child element instead of the XElement attribute during parsing.

Added a corresponding test case in the PackageReferenceFixture file.

*If it is a draft, describe the next steps
*Remember to create tests whenever possible

### DIFF
--- a/src/Buildalyzer/Construction/PackageReference.cs
+++ b/src/Buildalyzer/Construction/PackageReference.cs
@@ -12,6 +12,6 @@ public class PackageReference : IPackageReference
         this.Name = packageReferenceElement.GetAttributeValue("Include") ?? packageReferenceElement.GetAttributeValue("Update");
         var versionElement = packageReferenceElement.DescendantsAndSelf()
             .FirstOrDefault(x => x.Name.LocalName == "Version");
-        this.Version = versionElement?.Value ?? packageReferenceElement.GetAttributeValue("Version");
+        this.Version = packageReferenceElement.GetAttributeValue("Version") ?? versionElement?.Value;
     }
 }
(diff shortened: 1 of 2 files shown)


src/Buildalyzer/Construction/PackageReference.cs
```diff
@@ -10,6 +10,8 @@ public class PackageReference : IPackageReference
     internal PackageReference(XElement packageReferenceElement)
     {
         this.Name = packageReferenceElement.GetAttributeValue("Include") ?? packageReferenceElement.GetAttributeValue("Update");
-        this.Version = packageReferenceElement.GetAttributeValue("Version");
+        var versionElement = packageReferenceElement.DescendantsAndSelf()
+            .FirstOrDefault(x => x.Name.LocalName == "Version");
+        this.Version = versionElement?.Value ?? packageReferenceElement.GetAttributeValue("Version");
```
Corniel, line 15: Are you sure that the `<Version>x</Version>` overrules the `Version` attribute?

Corniel:
Can you try a rebase?

src/Buildalyzer/Construction/PackageReference.cs
Corniel: It feels weird (and wrong) for sure, but it should pick the same one (if they differ) as MS Build does. (and in case MS Build does return an error we might want to document that).

Corniel: It turns out attributes have precedence.. So these changes are fine.

Corniel (approved):
LGTM

PR #313, opened 2025-07-16
Add macOS back to OS matrix

slang25:
## Description

Adding back the macOS CI runner job, removed unsupported dotnet runtimes, and updated some of the GHA workflows.


Corniel (approved):
In my memory, we still had `netstandard2.0` targets. But as that is not longer true, this seems a perfect valid change. LGTM!

PR #324, opened 2025-11-25
Add release pipeline

badcel:
## Add Release pipeline

This would be the "idea" of the pipeline. I'm not sure if the building will work. Locally I got errors related to net6.0 which is not installed in the pipeline currently (but it is not installed in the already present build workflow, too).

The version number is defined in the `Directory.Build.props` file for reproducability. So if one checks the nuget package for the originating commit and does a `git checkout` for it, the produced nuget should be identical to the one from nuget.org. So you would need to bump the version before a new release.

The pipeline allows to define some "version suffix" like "preview.1" which would automatically result in a nuget package versioned for example like "7.1-preview.1".

I enabled [continuousintegrationbuild](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild) as I understand that it shoud be set for official releases.

I never used [nuget trusted publishing](https://learn.microsoft.com/en-us/nuget/nuget-org/trusted-publishing) myself (I still use api keys for my projects). But there is some configuration needed. See the linked documentation.

Fixes: #323

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,44 @@
+  name: 'Release'
+  
+  on:
+    workflow_dispatch:
+      inputs:
+        version_suffix:
+          description: Suffix of the version number. Can be used to create a preview package.
+          required: false
+          type: string
+
+  env:
+    configuration: Release
+    version_suffix: ${{ github.event.inputs.version_suffix }}
+  
+  jobs:
+    build-and-publish:
+      runs-on: 'ubuntu-latest'
+      permissions:
+        id-token: write  # enable GitHub OIDC token issuance for this job
+
+      steps:
+        - name: Get Source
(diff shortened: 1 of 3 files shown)


slang25:
I've just set up NuGet trusted publishing for dotnet-outdated and it was pretty straight forward. @phmonte would you be happy to add me as a package publisher so I can help with this?

slang25:
How about we publish a version to [feedz.io](https://feedz.io/) while we try to sort out the nuget.org credentials.

Corniel:
Is this one stale?

Corniel:
@badcel Time to have a look?

badcel:
Sorry I don't have the capacity to care for this PR anymore.

Feel free to take over the code.

PR #326, opened 2025-12-02
Support .NET 10

Corniel:
.NET 10 is out for almost a month now. It is time to support it.


slang25 (approved):
Nice one @Corniel

src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
```diff
@@ -3,7 +3,7 @@
   <Import Project="../../package.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net8.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net8.0;net10.0</TargetFrameworks>
```
Frulfump: Should the net6.0 TFM be de dropped here?

phmonte:
@Corniel  Is there any reason why we shouldn't update Buildalyzer for .NET 10 as well?

Corniel:
@phmonte Yes we should, but as a seperate story.

src/Buildalyzer.Workspaces/Buildalyzer.Workspaces.csproj
Corniel: Yes, but not here. Dropping a TFM is a breaking change. It's is planned, but postpone when we have other breaking changings.

PR #329, opened 2026-04-17
Fix binlog replay with MSBuild 18.x and normalize obsolete constructor path

mscottford:
## Summary

Three related fixes for reading binlogs produced by the .NET 10 SDK (MSBuild 18.x):

- **`AnalyzerManager.Analyze`**: replace `StructuredLogger.BinLogReader` with `Microsoft.Build.Logging.BinaryLogReplayEventSource`. The official MSBuild reader handles all 18.x event types (including `AssemblyLoadBuildEventArgs`), whereas `BinLogReader` from `MSBuild.StructuredLogger` may stop replaying mid-stream when it encounters unrecognised event types, leaving `_evalulationResults` empty so `ProjectStarted` falls back to `null` propertiesAndItems and produces no results.

- **`EventProcessor.ProjectStarted`**: treat an empty `e.Properties` collection the same as `null`. `BinLogReader` creates `ProjectStartedEventArgs` with `Properties` set to a non-null but empty collection for binlog format 14+ events, so the existing `e.Properties is null` check always fell through to the old-format path, using the empty collection instead of the populated evaluation-phase results. This left `ProjectReferences` empty after a binlog round-trip.

- **`AnalyzerManager` obsolete string constructor**: wrap the path in `Path.GetFullPath()` before passing it to `IOPath.Parse()`, restoring the behaviour of the previous `NormalizePath` implementation that was lost when `IOPath` was introduced.

Regression tests are included for the first two fixes.

## Test plan

- [ ] `SdkProjectWithProjectReferenceGetsReferencesFromBinaryLog` — builds a project with a project reference, replays the binlog, asserts `ProjectReferences` are populated
- [ ] `SdkProjectWithProjectReferenceGetsReferencesFromChildSideBinaryLog` — same but using child-side `/bl:` logging
- [ ] Existing integration test suite passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Buildalyzer/Logging/EventProcessor.cs
+++ b/src/Buildalyzer/Logging/EventProcessor.cs
@@ -83,8 +83,12 @@ internal class EventProcessor : IDisposable
         // Make sure this is the same project, nested MSBuild tasks may have spawned additional builds of other projects
         if (AnalyzerManager.NormalizePath(e.ProjectFile) == _projectFilePath)
         {
-            // Get the items and properties from the evaluation if needed
-            PropertiesAndItems propertiesAndItems = e.Properties is null
+            // Get the items and properties from the evaluation if needed.
+            // Treat an empty e.Properties the same as null: BinLogReader (MSBuild.StructuredLogger)
+            // creates ProjectStartedEventArgs with Properties set to a non-null but empty collection
+            // rather than null for binlog format 14+ events.  The empty collection carries no useful
+            // data, so fall back to the evaluation-phase results just as we do when Properties is null.
+            PropertiesAndItems propertiesAndItems = e.Properties is null || !e.Properties.Cast<object>().Any()
                 ? (_evalulationResults.TryGetValue(e.BuildEventContext.EvaluationId, out PropertiesAndItems evaluationResult)
                     ? evaluationResult
                     : null)
--- a/tests/Buildalyzer.Tests/Integration/SimpleProjectsFixture.cs
+++ b/tests/Buildalyzer.Tests/Integration/SimpleProjectsFixture.cs
@@ -374,6 +374,99 @@ public class SimpleProjectsFixture
         references.ShouldContain(x => x.EndsWith("SdkNetStandardProject.csproj"), log.ToString());
     }
 
+    /// <summary>
+    /// Regression test for the bug where ProjectReferences are empty when an analysis result
+    /// is obtained by replaying a binlog via <see cref="IAnalyzerManager.Analyze"/>.
+    ///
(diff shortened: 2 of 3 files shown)


src/Buildalyzer/Logging/EventProcessor.cs
```diff
@@ -83,8 +83,12 @@ private void ProjectStarted(object sender, ProjectStartedEventArgs e)
         // Make sure this is the same project, nested MSBuild tasks may have spawned additional builds of other projects
         if (AnalyzerManager.NormalizePath(e.ProjectFile) == _projectFilePath)
```
Corniel, line 84: For readablilty, I suggest you create an file
`Extensions/System.Collections.IEnumerable.cs`, containing te following:

``` csharp
namespace System.Collections;

internal static class EnumerableExtensions
{
    extension(IEnumerable? enumerable)
    {
        public bool HasNone
            => enumerable?.GetEnumerator().MoveNext() is not true;
    }
}
````

Than you don't need the explanation at all, als `e.Proparties.HasNone`, tells the story.

Note, that `<LangVersion>14</LangVersion>` has to be added, which is a good thing anyway.

File: tests/Buildalyzer.Tests/Integration/SimpleProjectsFixture.cs [hunk 2]
```diff
@@ -374,6 +374,99 @@ public void SdkProjectWithProjectReferenceGetsReferences()
...
+    /// back to evaluation-phase items. However, <c>BinLogReader</c> from
+    /// MSBuild.StructuredLogger creates <c>ProjectStartedEventArgs</c> with
+    /// <c>Properties</c> set to an empty non-null collection rather than <c>null</c>,
+    /// so the check always resolves to the old-format path — using the empty
+    /// <c>e.Properties</c>/<c>e.Items</c> instead of the populated evaluation results.
+    /// The fix is to treat an empty <c>e.Properties</c> the same as null.
+    /// </summary>
+    [Test]
+    public void SdkProjectWithProjectReferenceGetsReferencesFromBinaryLog(
+        [ValueSource(nameof(Preferences))] EnvironmentPreference preference)
```
Corniel, line 392: Use `TestTools.Context.ForProject`

Corniel (changes requested):
Nice improvements, and good spots! I'd like some (ecstatic) changes, but nothing to dramatic.

src/Buildalyzer/Logging/EventProcessor.cs
Corniel: After some consideration: Lets first merge #331 what does part the changes you want, and then follow with the rest.

PR #331, opened 2026-04-18
Factor out PropertiesAndItems.TryCreate

Corniel:
Looking at #329 I thought besides other changes, this logic can be cleaned up.

### DIFF
--- a/src/Buildalyzer/Extensions/System.Collections.IEnumerable.cs
+++ b/src/Buildalyzer/Extensions/System.Collections.IEnumerable.cs
@@ -4,6 +4,23 @@ internal static class EnumerableExtensions
 {
     extension(IEnumerable enumerable)
     {
-        public bool HasAny => enumerable.GetEnumerator().MoveNext();
+        public bool HasAny
+        {
+            get
+            {
+                var enumerator = enumerable.GetEnumerator();
+                try
+                {
+                    return enumerator.MoveNext();
+                }
+                finally
+                {
+                    if (enumerator is IDisposable disposable)
+                    {
+                        disposable.Dispose();
+                    }
+                }
+            }
+        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Buildalyzer/Extensions/System.Collections.IEnumerable.cs
```diff
@@ -0,0 +1,9 @@
+namespace System.Collections;
+
+internal static class EnumerableExtensions
+{
+    extension(IEnumerable enumerable)
+    {
+        public bool HasAny => enumerable.GetEnumerator().MoveNext();
```
slang25: Nit: I think this leaks the enumerator, we should dispose it

```suggestion
        public bool HasAny
        {
            get
            {
                var e = enumerable.GetEnumerator();
                try { return e.MoveNext(); }
                finally { (e as IDisposable)?.Dispose(); }
            }
        }
```

slang25: Or this could be this, which will lower to the same thing
```suggestion
        public bool HasAny
        {
            get
            {
                foreach (var _ in enumerable)
                {
                    return true;
                }
                return false;
            }
        }
```

Corniel: @slang25 I whent for your first option (with a twist). Although I think that the likelyhood of this iterator actually doing something while disposing is minimal.

Corniel:
@slang25 Packing fails. Do you have any experience with that?

slang25 (approved):

slang25:
> @slang25 Packing fails. Do you have any experience with that?

I don't, we need @phmonte for that

phmonte:
I can fix today!

Corniel:
> I can fix today!

That would be nice. I guess we should focus on #332 for that fix?! And maybe also on #324 (after that).

phmonte:
The pipeline that failed shouldn't have been triggered in the first place,  I'm resolving it.

phmonte:
Release published, @Corniel,I resolved the pipeline that was having problems and took the opportunity to document it [here](https://github.com/Buildalyzer/Buildalyzer/blob/main/CONTRIBUTING.md), the publishing process isn't great.

I'll take a look at #332 .

PR #337, opened 2026-04-19
Target .NET 10 with CPM

Corniel:
As we should not deploy breaking changes every day, I think we should not drop .NET 6 (yet).

- [x] .NET 10 targets
- [x] Enable `<EnforceCodeStyleInBuild>` (Roslyn's IDE rules) 
- [ ] Central Package Management
- [ ] Use lock files
- [x] Update packages with security Issues
- [x] Include PolySharp

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,36 +1,46 @@
 <Project>
 
-  <PropertyGroup>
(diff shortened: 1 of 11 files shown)


Corniel:
Discontinued

PR #338, opened 2026-04-20
Enforce coding style

Corniel:
Applied a bunch of IDE rules, to clean-up the code base.


src/Buildalyzer/Compiler/CompilerItems.cs
```diff
@@ -10,13 +10,11 @@ namespace Buildalyzer;
 [DebuggerTypeProxy(typeof(Diagnostics.CollectionDebugView<ITaskItem>))]
 public readonly struct CompilerItems(string key, IReadOnlyCollection<ITaskItem> values) : IReadOnlyCollection<ITaskItem>
 {
-    private readonly IReadOnlyCollection<ITaskItem> _values = values;
-
     /// <summary>Ghets the compiler item key.</summary>
```
slang25: Not your change, but while you're here
```suggestion
    /// <summary>Gets the compiler item key.</summary>
```

src/Buildalyzer/AnalyzerManager.cs
```diff
@@ -110,19 +110,13 @@ public IAnalyzerResults Analyze(string binLogPath, IEnumerable<Microsoft.Build.F
         };
     }
 
-    private IProjectAnalyzer? GetProject(IOPath path, ProjectInfo? project)
-    {
-        Guard.NotDefault(path);
-
-        if (!Guard.NotDefault(path).File()!.Exists)
+    private IProjectAnalyzer? GetProject(in IOPath path, ProjectInfo? project)
```
slang25: This goes for all `in IOPath` in this PR;

I find the `in` parameter on public methods a bit aesthetically weird. If it were on low level APIs or internal APIs then I'd be fine with it, but for prominent high-level methods like this, it should only exist if there's a real perf benefit.

I don't think there is a benefit here, because the struct has a single field, which is a string reference. So copying the struct would be the same cost as copying the reference to the struct, so perf wise it's exactly the same.

src/Buildalyzer/ProjectAnalyzer.cs
```diff
@@ -308,20 +307,16 @@ private string GetCommand(
...
-        if (string.IsNullOrEmpty(loggerDllPathEnv))
-        {
-            throw new ArgumentException($"The dll of {nameof(BuildalyzerLogger)} is required");
-        }
-
-        return loggerDllPathEnv;
-    }
+            { Length: > 0 } path => path,
+            _ when LoggerPathDll() is { Length: > 0 } path => path,
+            _ => throw new ArgumentException($"The dll of {nameof(BuildalyzerLogger)} is required"),
```
slang25: Not your change I know, but this should be a different exception type, `ArgumentException` are for argument preconditions,

tests/Buildalyzer.Tests/Compiler/CompilerCommandFixture.cs
```diff
@@ -123,43 +123,45 @@ public void Parse_VB()
...
+            --flaterrors
+            --highentropyva+
+            --targetprofile:netcore
+            --nocopyfsharpcore
+            --deterministic+
+            --simpleresolution
+            obj\Debug\netcoreapp3.1\.NETCoreApp,Version=v3.1.AssemblyAttributes.fs
+            obj\Debug\netcoreapp3.1\FSharpProject.AssemblyInfo.fs
+            Program.fs
+            """;
```
slang25, line 164: this makes me happy 😁

src/Buildalyzer/AnalyzerManager.cs
Corniel: @slang25 I guess you're right. I just blindly applied the suggestion. I'll roll it back.

src/Buildalyzer/ProjectAnalyzer.cs
Corniel: Yep. And I'm perfectly fine with change the Type here.

tests/Buildalyzer.Tests/Compiler/CompilerCommandFixture.cs
Corniel: Me too!

src/Buildalyzer/ProjectAnalyzer.cs
Corniel: I've changed it to `InvalidOperationException`

slang25 (approved):

PR #347, opened 2026-05-08
Fix stale XML doc on EnvironmentOptions.DesignTime (#343)

### LINKED ISSUE #343: Stale XML doc on EnvironmentOptions.DesignTime — TargetsToBuild is not actually ignored

## Symptom

The XML doc on `EnvironmentOptions.DesignTime` claims that `TargetsToBuild` is ignored when `DesignTime` is `true`:

```csharp
// src/Buildalyzer/Environment/EnvironmentOptions.cs:16-25
/// <summary>
/// Indicates that a design-time build should be performed.
/// The default value is <c>true</c>. Note that when performing
/// a design-time build, the <see cref="TargetsToBuild"/> will
/// be ignored and the design-time targets will be used instead.
/// </summary>
public bool DesignTime { get; set; } = true;
```

The implementation does not match. With the default `DesignTime = true` and the default `TargetsToBuild = ["Clean", "Build"]`, Buildalyzer issues `/t:Clean;Build` to MSBuild. Design-time *properties* are applied (e.g. `DesignTimeBuild=true`, `BuildingProject=false`, `SkipCompilerExecution=true` — these affect MSBuild item groups and the `Csc` task's behaviour), but the *targets list* passed in is whatever the caller set.

## Root cause

`EnvironmentFactory` and `BuildCommand` simply forward `options.TargetsToBuild` regardless of `options.DesignTime`:

```csharp
// src/Buildalyzer/Environment/EnvironmentFactory.cs:113
return new BuildEnvironment(
    options.DesignTime,
    options.Restore,
    [.. options.TargetsToBuild],   // ← passed through unchanged
    ...);
```

```csharp
// src/Buildalyzer/BuildCommand.cs:52
BuildArgument.Target(targetsToBuild is { Count: > 0 } ? targetsToBuild : env.TargetsToBuild),
```

There is no codepath that swaps `TargetsToBuild` for a design-time target list when `DesignTime` is true.

## Why it matters

The wording suggests a behaviour Buildalyzer doesn't have, which has two practical consequences:

1. Users assume Buildalyzer is invoking only design-time-safe targets and don't realise their `BeforeBuild`/`AfterBuild` hooks are firing.
2. Users who know about the `Clean;Build` cost don't realise that simply lowering `TargetsToBuild` to `["Compile"]` is the supported way to make Buildalyzer faster — they assume `DesignTime=true` already does that.

## Suggested fix

Either:

- **Update the doc** to reflect actual behaviour. Something like:

  > "Sets a number of MSBuild properties (`DesignTimeBuild=true`, `BuildingProject=false`, `SkipCompilerExecution=false`, `ProvideCommandLineArgs=true`, etc.) that adjust how the build behaves. **`TargetsToBuild` is still honoured.** Set `TargetsToBuild = ["Compile"]` for the lightest design-time evaluation."

- **Or change the behaviour** to actually substitute design-time targets when `DesignTime=true` — but this is a breaking change for anyone relying on the current behaviour.

The doc fix is the safe option. See also the related discussion of making `["Compile"]` the default in the companion issue I'm filing about `TargetsToBuild`.

## Verified against

- `src/Buildalyzer/Environment/EnvironmentOptions.cs:14,16-25`
- `src/Buildalyzer/Environment/EnvironmentFactory.cs:113,152`
- `src/Buildalyzer/BuildCommand.cs:52`

Related: #340.

slang25:
## Description
Fixes #343. The XML doc on `EnvironmentOptions.DesignTime` claimed that `TargetsToBuild` is ignored when `DesignTime` is `true`, but the implementation still forwards `TargetsToBuild` unchanged. This PR updates the doc to describe the actual behaviour: design-time mode sets MSBuild properties (`DesignTimeBuild=true`, `BuildingProject=false`, `SkipCompilerExecution=true`, `ProvideCommandLineArgs=true`) but `TargetsToBuild` is still honoured, with a hint that `TargetsToBuild = ["Compile"]` gives the lightest design-time evaluation.

### DIFF
--- a/src/Buildalyzer/Environment/EnvironmentOptions.cs
+++ b/src/Buildalyzer/Environment/EnvironmentOptions.cs
@@ -15,9 +15,12 @@ public class EnvironmentOptions
 
     /// <summary>
     /// Indicates that a design-time build should be performed.
-    /// The default value is <c>true</c>. Note that when performing
-    /// a design-time build, the <see cref="TargetsToBuild"/> will
-    /// be ignored and the design-time targets will be used instead.
+    /// The default value is <c>true</c>. Sets a number of MSBuild properties
+    /// (such as <c>DesignTimeBuild=true</c>, <c>BuildingProject=false</c>,
+    /// <c>SkipCompilerExecution=true</c>, and <c>ProvideCommandLineArgs=true</c>)
+    /// that adjust how the build behaves. <see cref="TargetsToBuild"/> is still
+    /// honoured; set <c>TargetsToBuild = ["Compile"]</c> for the lightest
+    /// design-time evaluation.
     /// </summary>
     /// <remarks>
     /// See https://github.com/dotnet/project-system/blob/master/docs/design-time-builds.md.


slang25:
Preferring #348

PR #348, opened 2026-05-08
Default TargetsToBuild to ["Compile"] (fixes #344)

### LINKED ISSUE #344: Lighter default TargetsToBuild for design-time / semantic-model use (30–50% cold-load speedup)

## Proposal

Add a documented opt-in (e.g. `EnvironmentOptions.MinimalTargets`) — or, more aggressively, change the default of `EnvironmentOptions.TargetsToBuild` from `["Clean", "Build"]` to `["Compile"]`.

## Context

`Build`'s dependency closure is broad:

```
BeforeBuild → PrepareForBuild → ResolveReferences →
GenerateAssemblyInfo → Compile → CoreCompile → AfterCompile →
CopyFilesToOutputDirectory → AfterBuild
```

`Compile` alone is everything Buildalyzer actually needs: `ResolveAssemblyReferences` runs, `Csc` fires its `TaskCommandLineEventArgs` (the event Buildalyzer parses to populate `CompilerCommand`), and evaluation-time `Items["Compile"]` is already captured at `ProjectStarted`. Anything past `CoreCompile` (`CopyFilesToOutputDirectory`, `AfterBuild`, etc.) does not contribute to the result Buildalyzer surfaces.

In an out-of-tree experiment (built around #340) we set:

```csharp
new EnvironmentOptions
{
    DesignTime = false,
    TargetsToBuild = { "Compile" }
}
```

(plus applying the design-time properties manually) and measured a **30–50% cold-load speedup** across 9 external projects (Newtonsoft.Json, Serilog, Polly, Humanizer, MediatR, FluentValidation, AutoFixture, NodaTime, StackExchange.Redis), with **identical document/reference counts** and identical `WorkspaceFingerprint` against `MSBuildWorkspace`. The win comes from skipping `BeforeBuild`/`AfterBuild` and the third-party hooks they reach (some of which are slow or fail-prone — e.g., NBGV's git-history walk).

## Implementation sketch

Two viable shapes:

1. **Behaviour change**: default `TargetsToBuild` to `["Compile"]`. This is technically breaking — users with `BeforeBuild`/`AfterBuild` hooks that they intentionally rely on will see them disappear. But given Buildalyzer is positioned as an evaluation/analysis tool not a build runner, those callers are arguably misusing the API.

2. **New option**: keep current default, add `EnvironmentOptions.MinimalTargets` (bool, default false) that substitutes `["Compile"]` and skips the `NonExistentFile` global property (which exists specifically to force `CoreCompile` to re-run under `Clean;Build`, see `EnvironmentFactory.cs:80-83`). Document the tradeoff. Non-breaking; opt-in speedup.

Option 2 is the safer first step.

## Related code

- `src/Buildalyzer/Environment/EnvironmentOptions.cs:14` — default `TargetsToBuild = ["Clean", "Build"]`.
- `src/Buildalyzer/Environment/EnvironmentFactory.cs:80-83,130-133` — the `NonExistentFile` workaround that's needed only with `Clean`.
- `src/Buildalyzer/BuildCommand.cs:52` — where the targets are forwarded to MSBuild.

## What this does *not* fix

Narrowing the target list does **not** bypass build-time tasks that hook in via `BeforeTargets` against targets inside `Compile`'s closure. In particular, Nerdbank.GitVersioning's `GetBuildVersion` still gets pulled in (it attaches to multiple targets) and still fails on shallow clones. That's the empty-workspace symptom in the companion issue (Csc-doesn't-run); same architectural root and needs a result-harvesting fix, not a target-list change.

## Verified against

- `src/Buildalyzer/Environment/EnvironmentOptions.cs:14`
- `src/Buildalyzer/Environment/EnvironmentFactory.cs:80-119`
- `src/Buildalyzer/BuildCommand.cs:45-58`

Related: #340.

slang25:
## Description

Changes the default `EnvironmentOptions.TargetsToBuild` from `["Clean", "Build"]` to `["Compile"]`, which mirrors the target Visual Studio's design-time builds drive (per the [project-system docs](https://github.com/dotnet/project-system/blob/main/docs/design-time-builds.md)) and what Roslyn's `MSBuildWorkspace` executes. `Compile`'s closure runs `ResolveReferences`, `GenerateAssemblyInfo` and `CoreCompile` (so `Csc` still emits the `TaskCommandLineEventArgs` Buildalyzer parses) but stops before `AfterCompile` / `CopyFilesToOutputDirectory` / `AfterBuild`, avoiding `BeforeBuild`/`AfterBuild` hooks and the third-party tasks attached to them.

Multi-targeted projects now schedule one inner build per TFM (matching how VS schedules design-time builds) instead of going through the outer `DispatchToInnerBuilds`, since `Compile` doesn't exist on the cross-targeting outer build. The empty-TFM "outer" result is no longer produced.

### WPF

`Compile` plus the default design-time properties still runs WPF's markup compilation: `Microsoft.WinFX.targets` hooks `DesignTimeMarkupCompilation` into `CoreCompileDependsOn` when `DesignTimeBuild` is set, so `MarkupCompilePass1` runs and XAML `*.g.cs` sources are surfaced — the same fidelity as a VS design-time build. A new SDK-style WPF test project covers this. Files generated only by the full `Build` closure (`GeneratedInternalTypeHelper.g.cs` from `MarkupCompilePass2`, which even VS's DTB doesn't run) require opting into `TargetsToBuild = ["Build"]`; the legacy WPF test does exactly that, and the caveat is documented on `EnvironmentOptions`.

### Supporting changes

- The `NonExistentFile` workaround is now unconditional (matching Roslyn's `MSBuildWorkspace`) so `CoreCompile` reliably re-runs regardless of the targets specified.

- When a `BinaryLogger` is attached, each per-TFM build writes a TFM-suffixed binlog (e.g. `project.net8.0.binlog`) so builds don't overwrite one another. Only the log-file segment of `BinaryLogger.Parameters` is rewritten (composite parameters like `ProjectImports=None;LogFile=x.binlog` are handled), with unit tests.
- `RELEASE.md` now flags the behavior changes as breaking for the next major version.

Rebased onto `main` (post-AwesomeAssertions migration) and squashed. Closes #344.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/Buildalyzer/Environment/EnvironmentOptions.cs
```diff
@@ -9,18 +9,33 @@ public class EnvironmentOptions
     public EnvironmentPreference Preference { get; set; } = EnvironmentPreference.Core;
 
     /// <summary>
-    /// The default targets to build.
+    /// The targets to build. Defaults to <c>["Build"]</c>. <c>Clean</c> was
+    /// previously the first default target, but it does not contribute to
```
Corniel: I don't think this belongs here. This should be something for the release notes.

Corniel:
Are there tests that can proof the effect of the changes?

src/Buildalyzer/ProjectAnalyzer.cs
```diff
@@ -137,13 +198,22 @@ public IAnalyzerResults Build(string targetFramework, BuildEnvironment buildEnvi
             []);
 
     /// <inheritdoc/>
-    public IAnalyzerResults Build() => Build((string)null);
+    public IAnalyzerResults Build() =>
+        ProjectFile.IsMultiTargeted
+            ? Build(ProjectFile.TargetFrameworks)
+            : Build((string)null);
```
Corniel:
```suggestion
            : Build((string?)null);
```

src/Buildalyzer/ProjectAnalyzer.cs
```diff
@@ -137,13 +198,22 @@ public IAnalyzerResults Build(string targetFramework, BuildEnvironment buildEnvi
...
+        ProjectFile.IsMultiTargeted
+            ? Build(ProjectFile.TargetFrameworks)
+            : Build((string)null);
 
     /// <inheritdoc/>
-    public IAnalyzerResults Build(EnvironmentOptions environmentOptions) => Build((string)null, environmentOptions);
+    public IAnalyzerResults Build(EnvironmentOptions environmentOptions) =>
+        ProjectFile.IsMultiTargeted
+            ? Build(ProjectFile.TargetFrameworks, environmentOptions)
+            : Build((string)null, environmentOptions);
```
Corniel:
```suggestion
            : Build((string?)null, environmentOptions);
```

tests/Buildalyzer.Tests/Environment/EnvironmentFactoryFixture.cs
```diff
@@ -1,11 +1,32 @@
 using Buildalyzer.Environment;
+using Buildalyzer.TestTools;
```
Corniel, line 2: I thought we moved to AwesomeAssertions?!

Corniel (approved):
Generally speaking a great improvement. Only some nit picking things.

slang25:
Rebased onto main and addressed the review feedback:

- Converted the new tests to AwesomeAssertions (branch predated #354) and applied the nullable cast suggestions.
- Moved the `TargetsToBuild` rationale out of the implementation comment into the XML docs and RELEASE.md, which now flags the behavior changes as breaking for the next major version.
- Re: tests proving the effect — added `Default_TargetsToBuild_is_Compile_only` and `GetBuildEnvironment_sets_NonExistentFile_under_default_targets`, unit tests for the per-TFM binlog path rewriting, and a new SDK-style WPF test project (`SdkWpfLibrary`) proving the default `Compile` build still surfaces XAML `*.g.cs` sources via `DesignTimeMarkupCompilation`. The whole integration suite also now runs under the new default.
- Two new behaviors while addressing this: restore now runs only on the first per-TFM build (one restore covers all frameworks in `<TargetFrameworks>`), and the binlog parameter rewriting only touches the log-file segment so composite parameters like `ProjectImports=None;LogFile=x.binlog` aren't mangled.

slang25:
Heads up on the Windows CI results from the previous push:

- `BuildsProject(SdkMultiTargetingProject)` failures were caused by the restore-once optimization I added — restoring with the `TargetFramework` global property pinned writes an assets file for that framework only, so the second TFM's build failed. Reverted in cee6cd7; restore runs per TFM as originally implemented.
- `Builds_DesignTime(Framework, LegacyFrameworkProjectWithReference)` fails with *"Cannot create a file when that file already exists"* — this is **pre-existing on `main`** (failing there since 2026-06-06, on the old `Clean;Build` default) and appears tied to the updated windows-latest runner image (VS 18 / MSBuild 18.7). Probably worth a separate issue; it isn't caused by or fixable in this PR.

PR #349, opened 2026-05-08
Apply evaluated compile properties to CSharpCompilationOptions (fixes #342)

### LINKED ISSUE #342: CreateCompilationOptions silently drops AllowUnsafe, CheckOverflow, Deterministic, Platform, WarningLevel

## Symptom

A workspace constructed from `IAnalyzerResult.GetWorkspace()` has a `CSharpCompilationOptions` that ignores most of the project's evaluated compile-related properties. Specifically:

- `AllowUnsafe` is always **false**, even when the project sets `<AllowUnsafeBlocks>true</AllowUnsafeBlocks>`.
- `CheckOverflow` is always **false**, even when `<CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>`.
- `Deterministic` is always **false**, even when `<Deterministic>true</...>`.
- `Platform` is always `AnyCpu`, even when `<PlatformTarget>` is set.
- `WarningLevel` is always **4** (Roslyn's hard-coded default), even though the .NET SDK defaults `WarningLevel` to the C# language version number (e.g. 9 on `net9.0`, 10 on `net10.0`) via `Microsoft.NET.Sdk`'s props chain.

These are particularly invisible in practice because most projects *don't* set the corresponding properties — but the moment a project sets one (often in `Directory.Build.props`), Buildalyzer-built workspaces disagree silently with both csc and Roslyn's MSBuildWorkspace. Code with `unsafe` blocks won't compile inside Buildalyzer's workspace; code that relies on `checked` overflow semantics will get the wrong analysis.

## Root cause

`AnalyzerResultExtensions.CreateCompilationOptions` constructs `CSharpCompilationOptions` with only `outputKind` and `nullableContextOptions`:

```csharp
// src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs:203-244
private static CompilationOptions CreateCompilationOptions(
    IAnalyzerResult analyzerResult, string languageName)
{
    string outputType = analyzerResult.GetProperty("OutputType");
    OutputKind? kind = null;
    switch (outputType) { ... }

    if (kind.HasValue)
    {
        if (languageName == LanguageNames.CSharp)
        {
            Enum.TryParse(
                analyzerResult.GetProperty("Nullable"),
                ignoreCase: true,
                out NullableContextOptions nullable);

            CompilationOptions CreateCSharpCompilationOptions() =>
                new CSharpCompilationOptions(kind.Value, nullableContextOptions: nullable);

            return CreateCSharpCompilationOptions();
        }
        ...
    }
    return null;
}
```

The remaining properties live on `analyzerResult.Properties` after evaluation completes (regardless of whether `Csc` later runs) — they just aren't read here.

## Suggested fix

Read each property and chain `.With…` calls on the constructed `CSharpCompilationOptions`. Sketch:

```csharp
var opts = new CSharpCompilationOptions(kind.Value,
    nullableContextOptions: nullable);

if (bool.TryParse(analyzerResult.GetProperty("AllowUnsafeBlocks"), out var allowUnsafe))
    opts = opts.WithAllowUnsafe(allowUnsafe);

if (bool.TryParse(analyzerResult.GetProperty("CheckForOverflowUnderflow"), out var checkOverflow))
    opts = opts.WithOverflowChecks(checkOverflow);

if (bool.TryParse(analyzerResult.GetProperty("Deterministic"), out var deterministic))
    opts = opts.WithDeterministic(deterministic);

var platform = analyzerResult.GetProperty("PlatformTarget")
            ?? analyzerResult.GetProperty("Platform");
if (Enum.TryParse<Platform>(platform, ignoreCase: true, out var platformValue))
    opts = opts.WithPlatform(platformValue);

if (int.TryParse(analyzerResult.GetProperty("WarningLevel"), out var warningLevel))
    opts = opts.WithWarningLevel(warningLevel);

return opts;
```

Notes:

- `Platform` from MSBuild is the *solution* platform (typically "AnyCPU"); the per-project compiler target is `PlatformTarget`. Read `PlatformTarget` first, fall back to `Platform`.
- `WarningLevel` is silently wrong on every modern-TFM project today because the SDK's default isn't 4. Even projects that don't touch the property are affected.

## Verified against

- `src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs:203-244`
- Property names match those evaluated by `Microsoft.NET.Sdk`'s `Microsoft.NET.Sdk.CSharp.props` chain.

Related: #340.

slang25:
## Description
Fixes #342. `AnalyzerResultExtensions.CreateCompilationOptions` previously constructed `CSharpCompilationOptions` with only `OutputKind` and `NullableContextOptions`, silently dropping `AllowUnsafe`, `CheckOverflow`, `Deterministic`, `Platform`, and `WarningLevel`. The fix reads `AllowUnsafeBlocks`, `CheckForOverflowUnderflow`, `Deterministic`, `PlatformTarget` (falling back to `Platform`), and `WarningLevel` from the evaluated analyzer result and applies them via the corresponding `With...` calls. Adds a `SdkCompilationOptionsProject` test fixture and a `CompilationOptionsRespectEvaluatedProperties` test asserting all five values flow through.

### DIFF
--- a/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
+++ b/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
@@ -75,6 +75,27 @@ public class ProjectAnalyzerExtensionsFixture
         compilationOptions.OutputKind.ShouldBe(OutputKind.DynamicallyLinkedLibrary, log.ToString());
     }
 
+    [Test]
+    public void CompilationOptionsRespectEvaluatedProperties()
+    {
+        // Given
+        SafeStringWriter log = new SafeStringWriter();
+        IProjectAnalyzer analyzer = GetProjectAnalyzer(@"projects\SdkCompilationOptionsProject\SdkCompilationOptionsProject.csproj", log);
+
+        // When
+        Workspace workspace = analyzer.GetWorkspace();
+        var compilationOptions = (Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions)workspace.CurrentSolution.Projects.First().CompilationOptions;
+
+        // Then
+        string logged = log.ToString();
+        logged.ShouldNotContain("Workspace failed");
+        compilationOptions.AllowUnsafe.ShouldBeTrue(logged);
+        compilationOptions.CheckOverflow.ShouldBeTrue(logged);
+        compilationOptions.Deterministic.ShouldBeTrue(logged);
+        compilationOptions.Platform.ShouldBe(Platform.X64, logged);
+        compilationOptions.WarningLevel.ShouldBe(7, logged);
+    }
+
     [TestCase(false, 1)]
     [TestCase(true, 3)]
     public void AddsProjectReferences(bool addProjectReferences, int totalProjects)
--- a/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
+++ b/tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
@@ -78,22 +78,19 @@ public class ProjectAnalyzerExtensionsFixture
     [Test]
     public void CompilationOptionsRespectEvaluatedProperties()
     {
-        // Given
-        SafeStringWriter log = new SafeStringWriter();
-        IProjectAnalyzer analyzer = GetProjectAnalyzer(@"projects\SdkCompilationOptionsProject\SdkCompilationOptionsProject.csproj", log);
+        using var ctx = Context.ForProject(@"SdkCompilationOptionsProject\SdkCompilationOptionsProject.csproj");
 
-        // When
-        Workspace workspace = analyzer.GetWorkspace();
+        using var workspace = ctx.Analyzer.GetWorkspace();
         var compilationOptions = (Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions)workspace.CurrentSolution.Projects.First().CompilationOptions;
 
-        // Then
-        string logged = log.ToString();
-        logged.ShouldNotContain("Workspace failed");
-        compilationOptions.AllowUnsafe.ShouldBeTrue(logged);
-        compilationOptions.CheckOverflow.ShouldBeTrue(logged);
-        compilationOptions.Deterministic.ShouldBeTrue(logged);
-        compilationOptions.Platform.ShouldBe(Platform.X64, logged);
-        compilationOptions.WarningLevel.ShouldBe(7, logged);
+        compilationOptions.Should().BeEquivalentTo(new
+        {
+            AllowUnsafe = true,
+            CheckOverflow = true,
+            Deterministic = true,
+            Platform = Platform.X64,
+            WarningLevel = 7,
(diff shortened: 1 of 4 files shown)


tests/projects/SdkCompilationOptionsProject/Class1.cs
```diff
@@ -0,0 +1,5 @@
+namespace SdkCompilationOptionsProject;
```
Corniel: Do we need a compile item? If so, don't we have a shared one we can reuse?

tests/projects/SdkCompilationOptionsProject/SdkCompilationOptionsProject.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
Corniel:
```suggestion
    <TargetFramework>net10.0</TargetFramework>
```

tests/projects/SdkCompilationOptionsProject/Class1.cs
slang25: Good point, we don't 😅  removed

tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -75,6 +75,27 @@ public void CreatesCompilationOptions()
         compilationOptions.OutputKind.ShouldBe(OutputKind.DynamicallyLinkedLibrary, log.ToString());
     }
 
+    [Test]
+    public void CompilationOptionsRespectEvaluatedProperties()
+    {
+        // Given
+        SafeStringWriter log = new SafeStringWriter();
+        IProjectAnalyzer analyzer = GetProjectAnalyzer(@"projects\SdkCompilationOptionsProject\SdkCompilationOptionsProject.csproj", log);
```
Corniel: No `using var ctx = Context.ForProject` here?

tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -75,6 +75,27 @@ public void CreatesCompilationOptions()
...
+
+        // When
+        Workspace workspace = analyzer.GetWorkspace();
+        var compilationOptions = (Microsoft.CodeAnalysis.CSharp.CSharpCompilationOptions)workspace.CurrentSolution.Projects.First().CompilationOptions;
+
+        // Then
+        string logged = log.ToString();
+        logged.ShouldNotContain("Workspace failed");
+        compilationOptions.AllowUnsafe.ShouldBeTrue(logged);
```
Corniel: with AwesomeAssertions:

``` C#
compilationOptions.Should().BeEquvilentTo(new
{
    AllowUnsafe = true,
    CheckOverflow = true,
    Deterministic = true,
    Platform = Platform.X64,
    WarningLevel = 7,
});
````

Corniel (approved):
LGTM (some nit picking again).

PR #351, opened 2026-05-31
Fix NullReferenceException in GetWorkspace() for .slnx solutions (#350)

### LINKED ISSUE #350: `System.NullReferenceException` on `AnalyzerManager.GetWorkspace()`
Given the following code:

```csharp
public static Workspace Load(string solutionFile)
{
    var path = IOPath.Parse(solutionFile);
    var manager = new AnalyzerManager(path);
    var workspace = manager.GetWorkspace();
    return workspace;
}
```

Results in a `NullReferenceException` on the following line of `GetWorkspace()`:

```csharp
List<ProjectInSolution> projectsInOrder = [.. manager.SolutionFile.ProjectsInOrder];
```

Here `SolutionFile` is `null`, because the underlying `AnalyzerManager.Solution.Reference` is of type `SolutionModel` and not of type `SolutionFile` (as the cast in `AnalyzerManager` would suggest). I didn't dig much deeper into this, but it might be related to the same changes that deprecated `AnalyzerManager.SolutionFile` in the first place. Perhaps the internal logic of `GetWorkSpace()` should not depend on that anymore.

slang25:
## Description
This PR fixes a `NullReferenceException` thrown by `AnalyzerManager.GetWorkspace()` when the manager was loaded from a `.slnx` solution (issue #350). The cause is that `GetWorkspace()` dereferenced the obsolete `AnalyzerManager.SolutionFile`, which is `null` for `.slnx` solutions because their underlying reference is a `SolutionModel` rather than a `Microsoft.Build.Construction.SolutionFile`. The fix orders projects using the non-obsolete `SolutionInfo.Projects` (populated for both `.sln` and `.slnx`) and removes the dependency on the deprecated API. A regression test (`LoadsSlnxSolution`) plus a `SingleProject.slnx` fixture were added, and the existing `.sln` test still passes.

Closes #350

### DIFF
--- a/src/Buildalyzer.Workspaces/AnalyzerManagerExtensions.cs
+++ b/src/Buildalyzer.Workspaces/AnalyzerManagerExtensions.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Buildalyzer.IO;
 using Microsoft.CodeAnalysis;
 using Microsoft.Extensions.Logging;
 namespace Buildalyzer.Workspaces;
@@ -36,9 +37,16 @@ public static class AnalyzerManagerExtensions
             Microsoft.CodeAnalysis.SolutionInfo solutionInfo = Microsoft.CodeAnalysis.SolutionInfo.Create(SolutionId.CreateNewId(), VersionStamp.Default, solutionPath);
             workspace.AddSolution(solutionInfo);
 
-            // Sort the projects so the order that they're added to the workspace is the same order as the solution
-            List<string> projectsInOrder = [.. solution.Projects.Select(p => p.Path.ToString())];
-            results = [.. results.OrderBy(p => projectsInOrder.FindIndex(g => g == p.ProjectFilePath))];
+            // Sort the projects so the order that they're added to the workspace is the same order as the solution.
+            // IOPath's equality/hash honour the file system's case sensitivity, so the lookup is robust across
+            // platforms; projects not found in the solution sort last.
+            Dictionary<IOPath, int> order = [];
+            for (int i = 0; i < solution.Projects.Length; i++)
+            {
+                order[solution.Projects[i].Path] = i;
+            }
+
+            results = [.. results.OrderBy(p => order.TryGetValue(IOPath.Parse(p.ProjectFilePath), out int index) ? index : int.MaxValue)];
         }
 
         // Add each result to the new workspace (sorted in solution order above, if we have a solution)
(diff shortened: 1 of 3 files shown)


src/Buildalyzer.Workspaces/AnalyzerManagerExtensions.cs
```diff
@@ -31,14 +31,22 @@ public static AdhocWorkspace GetWorkspace(this IAnalyzerManager manager)
 
         // Create a new workspace and add the solution (if there was one)
         AdhocWorkspace workspace = manager.CreateWorkspace();
-        if (!string.IsNullOrEmpty(manager.SolutionFilePath))
+        if (manager.Solution is { } solution)
         {
-            Microsoft.CodeAnalysis.SolutionInfo solutionInfo = Microsoft.CodeAnalysis.SolutionInfo.Create(SolutionId.CreateNewId(), VersionStamp.Default, manager.SolutionFilePath);
+            string solutionPath = solution.Path.ToString();
+            Microsoft.CodeAnalysis.SolutionInfo solutionInfo = Microsoft.CodeAnalysis.SolutionInfo.Create(SolutionId.CreateNewId(), VersionStamp.Default, solutionPath);
```
Corniel, line 37: Maybe alias with `MSSolutionInfo` ?

tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -41,6 +41,24 @@ public void LoadsSolution()
...
+        SafeStringWriter log = new SafeStringWriter();
+        AnalyzerManager manager = new AnalyzerManager(solutionPath, new AnalyzerManagerOptions { LogWriter = log });
+
+        // When
+        using var workspace = manager.GetWorkspace();
+
+        // Then
+        string logged = log.ToString();
+        logged.ShouldNotContain("Workspace failed");
+        workspace.CurrentSolution.FilePath.ShouldBe(solutionPath);
```
Corniel, line 58:
```suggestion
        workspace.CurrentSolution.FilePath.Should().Be(solutionPath);
```

tests/Buildalyzer.Workspaces.Tests/ProjectAnalyzerExtensionsFixture.cs
```diff
@@ -41,6 +41,24 @@ public void LoadsSolution()
...
+        workspace.CurrentSolution.Projects.ShouldContain(p => p.Name == "SdkNetStandardProject");
```
Corniel, line 59:
```suggestion
        workspace.CurrentSolution.Projects.Should().Contain(p => p.Name == "SdkNetStandardProject");
```

Corniel (approved):
LGTM.

slang25:
I'll follow up with a Shouldly → AwesomeAssertions PR

PR #354, opened 2026-06-06
Drop ShouldBe

Corniel:
Having two assertion frameworks is overkill. We go for AwesomeAssertions.


slang25 (approved):
You know that I'm the maintainer of Shouldly right? 😅
But yeah, we should have one, and the equivalent checks in AA are wanted here.

PR #355, opened 2026-06-06
Drop marvin

Corniel:
As far as I can tell, it of no use anymore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -295,4 +295,3 @@ tests/repos/net6.0
 tests/repos/netcoreapp3.1
 
 cache/
-marvin/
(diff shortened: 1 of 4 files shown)


daveaglick (approved):

daveaglick:
Yeah, Marvin definitely isn't needed anymore. It was just my little personal release tool (package, upload, set release notes in GitHub, etc.), but I don't see it being in the mix anymore. It's also not "production ready" so probably not great for a project I'm not maintaining anymore anyway.

On that point, you have no idea how wonderful it is to see this work continue without me ❤️

Corniel:
@daveaglick Well, may be, now you do not longer have to do all the heavy lifting, you might feel comfortable to contribute yourself now and then. (No pressure)

PR #356, opened 2026-06-10
Replace MsBuildPipeLogger with XenoAtom.MsBuildPipeLogger

slang25:
## Description
Replaces the unmaintained `MsBuildPipeLogger.Server`/`.Logger` packages with the single `XenoAtom.MsBuildPipeLogger` fork (1.0.1), updating namespaces (`MsBuildPipeLogger` → `XenoAtom.MsBuildPipeLogger`) across `Buildalyzer` and `Buildalyzer.Logger`. The custom `BuildalyzerLogger` is preserved and now compiles against the fork's bundled logger assembly (shipped as copy-to-output content) and is packed alongside `Buildalyzer.Logger.dll` as before. `Microsoft.Build.Utilities.Core` is bumped `14.3.0` → `17.10.46` to match the logger's MSBuild version and pick up the patched advisory (GHSA-w3q9-fxm7-j8fq), and the code adapts to the fork's nullable annotations. The redundant logger package references were dropped from the test projects since the DLL now flows transitively, and all 220 existing tests pass unchanged.


daveaglick:
FWIW the original MsBuildPipeLogger has been passed off to @xoofx too. I'm not sure if he's planning on continuing the fork or backporting changes back to the original, which might make a difference whether you want to update here or wait for changes to land back upstream.

slang25:
Thanks @daveaglick, I stumbled on the transfer of ownership and this new package today, so wasn't sure if one package was intended to supersede the other. There is nothing intentionally being "fixed" here, which is arguably the worst kind of change 😅  but I like dependencies to be maintained. I'll close this PR, but feel free to chime in too @xoofx

xoofx:
Hey @slang25, yes you can use `XenoAtom.MsBuildPipeLogger`, as @daveaglick graciously transferred the original project but I had already to make a separate project in order to push it earlier (and simplify NuGet ownership). I will probably archive the `MsBuildPipeLogger` and reroute it to XenoAtom. The XenoAtom version should include better edge cases for shutdown/errors and is also shipping the MSBuild logger differently so that it avoids runtime error in MSBuild when loading such logger.

xoofx:
As for your PR, you should not put the logger in the same folder than your app that interacts with MSBuild, otherwise it can create runtime errors in MSBuild that are difficult to investigate, showing problem like matching .NET DLL versions.

(Edit: Oh, reading more carefully, that's probably what you are already doing, so that should be good!)

slang25:
Superseded by #359, which takes this further — Buildalyzer no longer depends on MSBuild in-process at all (builds run out-of-process and stream events over the pipe as XenoAtom's version-independent PipeBuildEventArgs, so core has zero Microsoft.Build.* dependency). Closing in favour of that.

PR #358, opened 2026-07-16
Fix empty workspace when MSBuild fails before Csc runs (#341)

### LINKED ISSUE #341: Buildalyzer.Workspaces returns an empty workspace when MSBuild fails before Csc runs

## Symptom

`IAnalyzerResult.GetWorkspace()` returns an `AdhocWorkspace` containing the project but with **zero documents and zero metadata references**, despite the underlying evaluation having succeeded. The caller has no easy way to detect this — `Succeeded`/`OverallSuccess` may both be `true` (the top-level *evaluation* succeeded) or `false` (build failed) and the workspace surface looks the same shape either way.

Reproduces deterministically on, e.g., a shallow git clone of any project that uses Nerdbank.GitVersioning: NBGV's `GetBuildVersion` task throws in `BeforeCompile`'s dependency closure, MSBuild aborts before `CoreCompile` invokes `Csc`, and the resulting workspace is empty.

## Root cause

`EventProcessor` captures evaluation-time properties and items on `ProjectStarted` and stores them via `AnalyzerResult.ProcessProject`:

```csharp
// src/Buildalyzer/Logging/EventProcessor.cs:75-106
private void ProjectStarted(object sender, ProjectStartedEventArgs e)
{
    ...
    var propertiesAndItems = PropertiesAndItems.TryCreate(e, _evalulationResults);
    ...
    result.ProcessProject(propertiesAndItems);
}
```

Those properties and items end up on `IAnalyzerResult.Properties` and `IAnalyzerResult.Items` regardless of whether the build later fails.

But the workspace-building public surface (`IAnalyzerResult.SourceFiles`, `.References`, `.AnalyzerReferences`, `.PreprocessorSymbols`, `.AdditionalFiles`) is sourced exclusively from `CompilerCommand`:

```csharp
// src/Buildalyzer/AnalyzerResult.cs:68-83
public string[] SourceFiles =>
    CompilerCommand?.SourceFiles.Select(f => f.ToString()).ToArray() ?? [];

public string[] References =>
    CompilerCommand?.MetadataReferences.ToArray() ?? [];

public string[] AnalyzerReferences =>
    CompilerCommand?.AnalyzerReferences.Select(r => r.ToString()).ToArray() ?? [];
```

`CompilerCommand` is only populated when the `Csc` task fires its `TaskCommandLineEventArgs`:

```csharp
// src/Buildalyzer/Logging/EventProcessor.cs:149-152
if (e is TaskCommandLineEventArgs cmd && cmd.TaskName.IsMatch("Csc"))
{
    result.ProcessCscCommandLine(cmd.CommandLine, _targetStack.Any(...));
}
```

If `CoreCompile` never reaches the `Csc` invocation, `CompilerCommand` stays null, every accessor returns an empty array, and `AnalyzerResultExtensions.AddToWorkspace` happily produces an empty workspace from those empty arrays.

## Why MSBuildWorkspace doesn't hit this

Roslyn's MSBuildWorkspace BuildHost queries the in-process Build engine API directly for workspace state and streams it back, regardless of whether `Build()` returned true. Buildalyzer is reconstructing the workspace from build-event log output, so anything that gates on a specific later task firing is fragile.

## Suggested fix

When `CompilerCommand` is null, fall back to evaluation-time data on `result.Items` / `result.Properties`:

- **Documents**: `result.Items["Compile"]` (each `IProjectItem` has the source path as `ItemSpec` relative to the project directory).
- **Preprocessor symbols**: `result.GetProperty("DefineConstants")`, semicolon-split.
- **LangVersion**: already done via `result.GetProperty("LangVersion")` in `CreateParseOptions` (`AnalyzerResultExtensions.cs:166`).
- **Metadata references**: harder — `Csc`'s reference list is the resolved closure from `ResolveAssemblyReferences`. The closest evaluation-time approximation is `result.Items["Reference"]` plus `result.Items["ReferencePath"]` if the latter ran. A degraded reference set is better than zero (callers can still get source-level semantic models).

Recommended shape: a private `BuildFromItems(...)` path inside `AnalyzerResultExtensions.GetDocuments` / `AnalyzerResultExtensions.GetMetadataReferences` that activates only when the `CompilerCommand`-backed path returns empty *and* `result.Items["Compile"]` is non-empty. We ship this as an out-of-tree workaround in a playground built around #340 (the "items-fallback" mode) — it recovers ~99% of documents on shallow-clone NBGV failures and is visibly correct when measured against MSBuildWorkspace.

## Workaround consumers should know about today

Always check the workspace contents after `GetWorkspace()`:

```csharp
var workspace = result.GetWorkspace();
var project = workspace.CurrentSolution.Projects.Single();
if (!project.Documents.Any())
{
    // Don't trust this workspace.
}
```

`result.Items["Compile"]` and `result.Properties` are still populated and can be used to construct a workspace directly.

## Verified against

- `src/Buildalyzer/Logging/EventProcessor.cs` (line refs above)
- `src/Buildalyzer/AnalyzerResult.cs` (line refs above)
- `src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs:259-296`

Related: #340.

slang25:
## Description
Fixes #341: `GetWorkspace()` returned a project with zero documents and references when MSBuild aborts before `CoreCompile` invokes the compiler (e.g. a shallow-clone Nerdbank.GitVersioning failure), because `CompilerCommand` is never captured. Buildalyzer.Workspaces now reconstructs workspace state from evaluation-time items/properties as a fallback — documents from `Compile`, preprocessor symbols from `DefineConstants`, metadata references from `ReferencePath`/`Reference`, analyzer references from `Analyzer`, and additional documents from `AdditionalFiles`. The fallback only activates when the compiler-backed data is empty and `Compile` items exist, so successful-build behavior is unchanged. Added a deterministic `FailBeforeCompile` test project (a target erroring `BeforeTargets="CoreCompile"`) plus tests asserting documents and preprocessor symbols are recovered.

### DIFF
--- a/src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
+++ b/src/Buildalyzer.Workspaces/AnalyzerResultExtensions.cs
@@ -160,7 +160,7 @@ public static class AnalyzerResultExtensions
                 CSharpParseOptions parseOptions = new CSharpParseOptions();
 
                 // Add any constants
-                parseOptions = parseOptions.WithPreprocessorSymbols(analyzerResult.PreprocessorSymbols);
+                parseOptions = parseOptions.WithPreprocessorSymbols(GetPreprocessorSymbols(analyzerResult));
 
                 // Get language version
                 string langVersion = analyzerResult.GetProperty("LangVersion");
@@ -294,6 +294,15 @@ public static class AnalyzerResultExtensions
     private static IEnumerable<DocumentInfo> GetDocuments(IAnalyzerResult analyzerResult, ProjectId projectId)
     {
         string[] sourceFiles = analyzerResult.SourceFiles ?? [];
+
+        // If MSBuild failed before the compiler ran, CompilerCommand is null and there are
+        // no source files. Fall back to the evaluation-time Compile items so the workspace
+        // still contains documents (see https://github.com/Buildalyzer/Buildalyzer/issues/341).
+        if (sourceFiles.Length == 0 && ShouldFallBackToItems(analyzerResult))
+        {
+            sourceFiles = GetItemPaths(analyzerResult, "Compile");
+        }
+
         return GetDocuments(sourceFiles, projectId);
     }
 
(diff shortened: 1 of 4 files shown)


slang25:
Superseded by the #341 fix incorporated into #356 (commit 2deab18 on slang25/update-buildalyzer-pr-356).

The approach is the same idea — reconstruct the workspace from evaluation-time items when MSBuild aborts before `CoreCompile` — but re-authored for #356's out-of-process architecture. That architecture doesn't have `ReferencePath` in the evaluation snapshot (it's produced by RAR during the build, not at evaluation), so the out-of-process logger now also forwards `ResolveAssemblyReference`'s resolved `ReferencePath` output; documents come from `Compile`, analyzers from `Analyzer`, additional docs from `AdditionalFiles`, and symbols from `DefineConstants`. Net result matches this PR (documents + references + analyzers + symbols recovered), verified by a fail-before-`CoreCompile` test.

Closing in favour of #356.

